PR #153, opened 2014-11-14
Call cpio with --extract-over-symlinks, if present

marcus-h:
This is needed to do a local build against the Fedora:20 project:
Fedora 20 ships a filesystem rpm that symlinks /lib64 to /usr/lib64. Moreover, the libgcc rpm,
for example, ships a "/lib64/libgcc_s-4.8.2-20131017.so.1". Thus, openSUSE's cpio cannot
extract the archive (except --extract-over-symlinks is specified).

### DIFF
--- a/init_buildsystem
+++ b/init_buildsystem
@@ -203,6 +203,7 @@ preinstall() {
     else
 	unsafe_preinstall_check
 	CPIO="cpio --extract --unconditional --preserve-modification-time --make-directories --no-absolute-filenames --quiet"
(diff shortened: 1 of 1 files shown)


adrianschroeter:
okay, closing it. (bsdtar had just a similar bug, but luckily fixed meanwhile ;)

rzr:
Hi,
Someone contacted me about this remaining issue, which version of gbs fixes this bug ?
https://bugs.tizen.org/jira/browse/DEVT-241

mlschroe:
IIRC: we won't add --extract-over-symlinks due to security implications, The "fix" is to install bsdtar as it supports a --chroot option that cpio is missing.

PR #155, opened 2015-01-14
build-vm-kvm: Switch armv7 to virt machine

dirkmueller:
This removes the need for DT and allows building a much smaller
guest kernel.


adrianschroeter:
merged

PR #156, opened 2015-01-14
ppc64le can build ppc64 with KVM

k0da:
Little Endian host can do build in Big Endian VM

Signed-off-by: Dinar Valeev <dvaleev@suse.com>


adrianschroeter:
merged

PR #158, opened 2015-02-24
Fixing UTC symlink resolving when setting localtime

JLahti:
The readlink command substitution should be escaped. Otherwise we are getting the path in the host env instead of the chroot env.


adrianschroeter:
right, thanks a lot

PR #159, opened 2015-03-02
Avoid emulator for arm on aarch64

andreas-schwab:


adrianschroeter:
are you sure that the arm* architectures can be build directly on an aarch64 system? don't you need kvm or alike for that?

andreas-schwab:
Works fine here, though you need to run with linux32 to get uname right.  KVM builds are using the aarch64 kernel too.

PR #160, opened 2015-03-03
Don't bin existing BUILD dir on --no-init.

thorstenb:
Permits incremental building inside buildroot.

### DIFF
--- a/build-recipe-arch
+++ b/build-recipe-arch
@@ -23,7 +23,9 @@
 
 recipe_setup_arch() {
     TOPDIR=/usr/src/packages
-    rm -rf "$BUILD_ROOT$TOPDIR"
+    if test "$DO_INIT" = true ; then
+	rm -rf "$BUILD_ROOT$TOPDIR"
+    fi
     mkdir -p "$BUILD_ROOT$TOPDIR"
     mkdir -p "$BUILD_ROOT$TOPDIR/OTHER"
     mkdir -p "$BUILD_ROOT$TOPDIR/SOURCES"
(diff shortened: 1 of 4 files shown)


adrianschroeter:
What exactly want you achieve here? rpmbuild would still remove this directory usually, no?

IMHO we should implement some "build --shell" option instead to jump inside. The user can call rpmbuild (or any other tool) directly with the stage options.

marcus-h:
On 2015-03-03 00:12:13 -0800, Adrian Schr�ter wrote:
> What exactly want you achieve here? rpmbuild would still remove this directory usually, no?
> 
I think something like this makes sense if you want to examine e.g. build
failures in the build chroot/vm (using "build ... --shell" or
"osc build ... --shell").

> IMHO we should implement some "build --shell" option instead to jump inside. The user can call rpmbuild (or any other tool) directly with the stage options.
> 
The build script already has such an option (but it indirectly removes
the TOPDIR).

thorstenb:
Sure, getting me a shell inside the build vm/chroot would permit that as well. Just that osc chroot does not currently seem to yield a functioning build environment.
But this patch is rather minimal, and enables me playing %setup -D -T 0 tricks. Otherwise fixing rpmlint problems in large projects (libreoffice, ceph) are pretty painful.

mlschroe:
I changed the code so that it does not erase $TOPDIR if --no-init is given twice. That way we stay compatible to old versions.

thorstenb:
Ah cool, that works. Thanks!

PR #162, opened 2015-03-05
Fix support for building inside LXC containers.

OlegGirko:
This change adds proper support for building inside lightweight
LXC containers, improving security without using heavyweight virtual machine.

Now LXC container is properly created on startup and destroyed on cleanup.
No need to destroy container when killing build process,
cleanup is following anyway.

Now LXC container has its private rootfs distinct from buld root.
Build root is bind mounted on LXC container's rootfs
inside container's namespace during container startup
(and unmounted automatically when build inside container finished).
This is needed to prevent destruction of build root by lxc-destroy
which removes container's rootfs recursively.

Logging has been fixed for LXC.
LXC is a special case because virtualised process shares its build root
and log file with host system,
so a check has been added to prevent re-creating log file by second stage.
Also, LXC container had to be specially configured to log its console output
to stdout of lxc-start process.

Also, container name is now prefixed with "obsbuild:" to prevent
name clashes with other LXC containers on the same host.

Signed-off-by: Oleg Girko <ol@infoserver.lv>

### DIFF
--- a/build
+++ b/build
@@ -980,7 +980,10 @@ if test -w /root ; then
 fi
 
 if test -z "$VM_IMAGE" -a -z "$LOGFILE" ; then
-    LOGFILE="$BUILD_ROOT/.build.log"
+    # lxc is special case: virtual machine shares logfile with host
+    if test -z "$RUNNING_IN_VM" -o "$VM_TYPE" != lxc; then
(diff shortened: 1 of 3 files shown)


adrianschroeter:
great, thanks for taking care of lxc! :)

PR #163, opened 2015-03-18
Add initial support for building inside Docker containers.

OlegGirko:
This change adds support for building inside lightweight Docker containers.

The following approach is used.
* Minimal busybox pre-built image is used
(downloaded from Docker Hub automatically on the first use).
* Docker container is started from this image.
* The container is configured with sys_admin capability granted
to allow second stage to mount "/proc" and "/sys".
* Build root is configured as a volume for this container, mounted at
"/mnt" directory inside the container.
* Second stage is started inside the container, chrooted in "/mnt" directory.

This convoluted approach is used for the following reasons.
* There's no way in Docker to mount container's root fs and populate it
(as with KVM build).
* You can only create containers from images, mount some host directories
to container's directories upon container creation
(no way to reconfigure these mounts later);
all further changes to container's filesystem can be done only by
running commands inside the container.
* Mounting build root as container's root (as with LXC build) is impossible.
There's empty pre-built image called scratch, but using it to create
container with build root directory mounted on container's root doesn't work:
scratch image can be used only as base for creating other images,
but not containers;
also, it's not allowed to mount host directories on container's root anyway.
* It's also possible to create an image for each build, but it's
too slow and space consuming: you have to populate build root,
then use its tar archive to build a new container,
and only after that create a container from this image
and run second stage inside it.
Also, retrieving build results from this container can be tricky.
Using Docker's lightweight layered image features is an idea
for further research.

### DIFF
--- a/build
+++ b/build
@@ -283,7 +283,7 @@ Known Parameters:
 
   --vm-type TYPE
               Use virtual machine instead of chroot
-              TYPE is one of xen|kvm|uml|qemu|lxc|zvm|openstack|ec2
+              TYPE is one of xen|kvm|uml|qemu|lxc|zvm|openstack|ec2|docker
 
   --vm-worker GUEST
               GUEST is a z/VM build worker controlled by the controlling
@@ -980,9 +980,15 @@ if test -w /root ; then
 fi
 
 if test -z "$VM_IMAGE" -a -z "$LOGFILE" ; then
-    # lxc is special case: virtual machine shares logfile with host
-    if test -z "$RUNNING_IN_VM" -o "$VM_TYPE" != lxc; then
-        LOGFILE="$BUILD_ROOT/.build.log"
+    # lxc and docker are special cases: virtual machine shares logfile with host
+    if test -z "$RUNNING_IN_VM"; then
+        case "$VM_TYPE" in
+        lxc|docker)
+            ;;
+        *)
+            LOGFILE="$BUILD_ROOT/.build.log"
+            ;;
+        esac
     fi
 fi
 
--- a/build-vm
+++ b/build-vm
@@ -60,7 +60,7 @@ HUGETLBFSPATH=
 # emulator specific?
 EMULATOR_SCRIPT=
 
-for i in ec2 emulator kvm lxc openstack qemu uml xen zvm ; do
+for i in ec2 emulator kvm lxc openstack qemu uml xen zvm docker ; do
     . "$BUILD_DIR/build-vm-$i"
 done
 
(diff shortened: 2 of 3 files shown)


adrianschroeter:
Understanding your description this is basically a variation of the LXC support with more limitations.
What is the advantage of using docker instead of LXC here?

Or is this just the ground work for future advantages?

The code is fine though, I am find with merging it. Just it would be nice if the busybox image could be provided directly. Otherwise it won't be possible to run this in an isolated network, like many OBS people do for additional security.

adrianschroeter:
Won't you want add also your name to the copyright to the new file?

OlegGirko:
Yes, this is essentially proof of concept that Docker can be used in place of LXC for this task.

There are some advantages of this approach over LXC.
* Its simplicity: temporary container is created and started with one command, no need to create config file and hook script, no cleanup necessary after task finished.
* No annoying error message when build inside the container finished with non-zero exit code.
* More security: no root privileges, just additional capability to mount `/proc` and `/sys`, and even this capability may be made unnecessary in future if pre-built image is used (see below).

There are, however, disadvantages of this approach comparing to LXC.
* Additional setup steps are necessary:
  * starting Docker service;
  * building busybox image for isolated environment (can be made unnecessary in future).
* Additional SELinux rules are necessary to prevent AVC denials (LXC works without additional rules with my OBS port to Fedora with SELinux enforcing mode).

This is an initial simple implementation that can probably be extended in future. Docker's layered image features can be used to speed up container creation. For example, container can be created from pre-built image that already contains all packages required by prjconf. This image can be reused for building all packages for the repo and rebuilt only when prjconf is changed or packages required by prjconf are updated. However, this requires changes to build and OBS.

I don't see a reason to include pre-built busybox image in build source. There are no image files in Docker, they are stored in some kind of internal storage by Docker service. It's possible to dump image, but this dump is just a tarball containing config files and tarballs for image layers. Anyway, additional setup steps are necessary to run build with Docker, so setup instructions can be included with steps to re-create busybox image. And this may become unnecessary in future anyway.

OlegGirko:
I've just force pushed a new version of my change.
Differences comparing to the previous version:
* added my name to copyright header of the new file;
* added `--net=none` option to `docker run` command to confine created container to isolated localhost-only network.

adrianschroeter:
Can you add the check for the remount call?

OlegGirko:
It took some time for me to understand data flow in build well enough to find that no vm specific hooks run inside virtual machine. All these hooks run only if `"$RUNNING_IN_VM"` is empty string.

Also, I found no easy way to find whether remount is necessary. There may me several entries for sysfs in `/proc/self/mounts` inside chroot in Docker container because there is sysfs already mounted outside chroot. These entries look the same.

On the other hand, this remount is pretty harmless, it just follows mount that already mounts sysfs rw. Unfortunately, sometimes, probably due to some kind of bug, sysfs is mounted ro anyway, and remount fixes it. If sysfs is mounted rw already, this remount does nothing.

PR #166, opened 2015-04-17
Support for simple images

miska:
Simple image is a really simple way how to get rootfs tarball or
squashfs image. The main advantage over existing solution like kiwi is
that it is distribution independent and therefore any distribution
supported by obs can make use of it. Disadvantage is that it is simple
and basically just preconfigured tarball - no bootloader, no kernel.

Syntax is pretty simple, all you need is to create simpleimage file and
write what you want inside of it. Here is the simple example, all fields
are optional as is the %build section.

            +----------------------------------+
            | Name:           example          |
            | Version:        1.0              |
            | BuildRequire:   emacs            |
            | #!BuildIgnore:  gcc-c++          |
            |                                  |
            | %build                           |
            | passwd << EOF                    |
            | opensuse                         |
            | opensuse                         |
            | EOF                              |
            +----------------------------------+

TODO: Figure out how to eliminate inherited dependencies.

### DIFF
--- a/Build.pm
+++ b/Build.pm
@@ -24,6 +24,7 @@ use strict;
 use Digest::MD5;
 use Build::Rpm;
 use Data::Dumper;
+use POSIX qw(strftime);
 
 our $expand_dbg;
 
@@ -1174,6 +1175,7 @@ sub recipe2buildtype {
   $recipe =~ s/^_service:.*://;
   return 'arch' if $recipe eq 'PKGBUILD';
   return 'preinstallimage' if $recipe eq '_preinstallimage';
+  return 'simpleimage' if $recipe eq 'simpleimage';
   return undef;
 }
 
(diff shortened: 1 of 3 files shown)


adrianschroeter:
Is "simpleimage" a standard somewhere or would this be an OBS special feature?

miska:
OBS special.

adrianschroeter:
Can you provide a documentation snipped for our book please describing what is it's purpose?
A use case when to use the simpleimage instead of existing kiwi/liveimage/preinstall images would be good as well.

It is enough to provide the text, I will turn it into the susebook xml afterwards. thanks.

miska:
Not sure whether it is in correct section or whether it compiles cleanly, but here it is:
https://github.com/openSUSE/obs-docu/pull/12

PR #171, opened 2015-04-30
Keep the debootstrapdir around so we can examine the log if things go wr...

jpwhiting:
...ong


adrianschroeter:
maybe we should also put out the log on stdout in addition, it would at least help a lot on server side failures?

jpwhiting:
Adrianschroeter, that's what 50337b63cd4881d992e67109fa1603b92ed6eb2c does, no?

PR #177, opened 2015-05-21
Force cpio(1) to extract files into symlinked directories

steelman:
A vulnerability[1] in cpio(1) has been reported in cpio which enables
directory traversal via symlinks. It has been patched and now
preinstalling RPM packages "over" the filesystem package which provides
/lib as a symlink to /usr/lib requires --extract-over-symlinks.

[1] https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-1197

### DIFF
--- a/init_buildsystem
+++ b/init_buildsystem
@@ -202,7 +202,7 @@ preinstall() {
 	TAR="/usr/bin/bsdtar -P --chroot -o --numeric-owner -x"
     else
 	unsafe_preinstall_check
-	CPIO="cpio --extract --unconditional --preserve-modification-time --make-directories --no-absolute-filenames --quiet"
+	CPIO="cpio --extract --unconditional --preserve-modification-time --make-directories --no-absolute-filenames --quiet --extract-over-symlinks"
 	TAR="tar -x"
     fi
     pkg_preinstall


adrianschroeter:
yes, cpio was never considered secure from our POV, therefore we defaulted to bsdtar (turned out it had also a security issue though). 

I assume that this patch will break support with old/unsecure cpio versions? Well, maybe a good thing, but we should show a proper error message then IMHO

marcus-h:
On 2015-06-10 02:40:35 -0700, Adrian Schröter wrote:
> I assume that this patch will break support with old/unsecure cpio versions?

Yes, this patch will break "old" cpio versions. Some time ago, we discussed
the pros and cons for "--extract-over-symlinks" in
https://github.com/openSUSE/obs-build/pull/153

steelman:
The main "pro" of the switch is that it makes build *work* with new cpio.

adrianschroeter:
it is not a regression when the new one is not working, but it is one when the old one does suddenly not work anymore ....

And as marcus-h reminded me, this would also hide errors in our packages we actually like to see. Having binaries in rpms which only work via symlinks creates the problem that also the preinstall dependencies would need to be checked, otherwise you can run into un-installable systems. Except you work again with %pre scripts to workaround the broken packaging.

So, we consider Fedora packaging currently as broken, but you can still use OBS and the build script when you install the recommended bsdtar.

PR #182, opened 2015-06-16
debian: lintian should check generated binary and source package

jblunck:
Since commit 7e49410794f58316b3afaf5cf78c5ddc2fba9def the linitan call was
only covering the source Debian package. Since the build architecture is
gathered from the chroot it should be safe to use the changes file as
lintian input.


HurricaneHernandez:
I believe this merge causes the following issue:https://bugzilla.novell.com/show_bug.cgi?id=939398

adrianschroeter:
@jblunck why does it fail after reporting just warnings? shouldn't it be only fail on errors?

jnweiger:
My debian builds work fine now. Thanks!

PR #191, opened 2015-07-30
Add the dpkg way of noscripts removal

gollub:
Required since --force all doesn't prevent that post/pre remove
scripts get called. Required since init_buildsystem is removing
package without honoring dependencies. (e.g. libexpt1 get removed
before python3 => python3 prerm scripts fail. Deinstallation fails.
osc build fails.)

### DIFF
--- a/build-pkg-deb
+++ b/build-pkg-deb
@@ -68,8 +68,22 @@ pkg_install_deb() {
 pkg_erase_deb() {
     export DEBIAN_FRONTEND=noninteractive
     export DEBIAN_PRIORITY=critical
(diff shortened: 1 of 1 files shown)


bluca:
:+1:

gollub:
@mlschroe , @adrianschroeter , do you want me to do further changes on this pull-request?

gollub:
@mlschroe , @adrianschroeter any review remarks?

gollub:
@adrianschroeter, @mlschroe , ping?

DStape:
@adrianschroeter, @mlschroe - can this be merged in?

adrianschroeter:
sorry for taking so long ... reason is that we really don't like this implementation, but we have no better ideas ....

So I am going to merge it :/

PR #195, opened 2015-08-24
Issue 884 failure timestamp

bgeuken:


bgeuken:
I have to admit that I wasn't able to test this one, because I wasn't able to setup the backend on my dev system.

PR #198, opened 2015-09-05
Fix auto-detection of -p1 option to patch

aspiers:
Check if either the `--- ` filename starts with `a/` or the `+++ ` filename starts with `b/`, or either starts with a package name/version prefix. We have to check for either, because either of them could be `/dev/null` if a file is being added or deleted.

### DIFF
--- a/spec_add_patch
+++ b/spec_add_patch
@@ -138,8 +138,16 @@ for my $diffname (keys %diffs) {
     my $striplevel = "";
     open(P, '<', $diffname) or die "$diffname: $!\n";
     while(<P>) {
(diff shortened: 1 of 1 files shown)


rhafer:
:+1:

rsalevsky:
+1

aplanas:
@mlschroe can you review this?

For me looks Perl, except for the comment. Maybe you want to remove the comment.

aspiers:
@aplanas Why are you suggesting to remove the comment?  It's necessary in order to understand why just checking for `/^--- a/` is not good enough.

@dirkmueller wrote this script so should probably give the final +2.

PR #194, opened 2015-08-10
Logging and debtransform improvements

jengelh:

PR #200, opened 2015-09-06
Bring back printing progress per package when initializing build root

jengelh:
(This is a repost of the one unmerged commit from pull request #194 so it does not get lost.)

This reverts commit e2a3a2681057293a452938cb523783f8d88809e6.

That commit crippled progress reporting. Issue 1: It made the
"keeping" messages go away, but the progress indicator would still be
printed, so the commit only did half of the intent.

Issue 2: Since no newline was emitted (this was part of the "keeping"
message) and because stdout is line-buffered, the progress report
would only be sporadically emitted.

Furthermore, initializating a Debian root is slow compared to RPM, and
so there is a long time before any output is shown.

Show me my progress as it happens. Reinstate messages about kept
packages.


adrianschroeter:
@lnussel can you comment on this please?

lnussel:
sorry, I missed the notification. I guess I didn't like the excessive output back then. Nowadays I don't care.

PR #202, opened 2015-09-15
Run initscripts during debian setup

chas3:
Building some packages have dependencies related to auto testing.
We need initscripts to be run so that the startup scripts are correctly
linked for later use by postinst scripts from packages like cgmanager.


gollub:
:+1:

bluca:
:+1:

gollub:
@mlschroe any chance to get this merged upstream? Or do you see any risk for Debian builds here?

adrianschroeter:
might make some errors visible for old environments, but it will not abort on failure. so it is fine IMHO.

PR #203, opened 2015-09-28
Fix finding patch Number for certain spec files

rhafer:
When a spec file has interleaved "Source" and "Patch" tags, the script
generated the wrong patchnumber (0) and put the patch after the last "Source"
line instead of after the last "Patch".


rhafer:
@dirkmueller maybe you want to review this. AFAIK you created this script initially.

dirkmueller:
+1

PR #204, opened 2015-10-10
Continuously output messages from rpmlint

StefanBruens:
Preparational patch for some rpmlint changes. For example for kernel builds, rpmlint may take dozens of minutes until any output appears. To at least allow rpmlint to report progress, use tee to continuously pipe the log to stdout.


StefanBruens:
Ping?

seife:
This will make ret=1 only if tee fails, not if rpmlint fails.

try:

false|tee /dev/null ||echo hallo

it will return nothing

StefanBruens:
Possible solutions:
false | tee > /dev/null  ; ret=${PIPESTATUS[0]} ; echo $ret
(set -o pipefail; false | tee > /dev/null ); ret=$? ; echo $ret

adrianschroeter:
I have added "set -e -o pipefail" here now. Should solve this.

PR #206, opened 2015-10-16
Skip empty packages in mkbaselibs.

PavelKopyl:
When building packages on a private OBS instance occasionally I get the following error: "Can't use an undefined value as an ARRAY reference at /usr/lib/build/mkbaselibs". It happens if mkbaselibs gets empty packages (for instance *.debuginfo.rpm) among input arguments. It seems reasonable to skip such packages.


adrianschroeter:
thanks! this was really anoying sometimes

PR #207, opened 2015-10-19
Add virtio-rng device to a VM

k0da:
In case we have /dev/hwrng present, use it as a backend device for
virtio-rng, fallback to /dev/random if hw is not there

Signed-off-by: Dinar Valeev <dvaleev@suse.com>

### DIFF
--- a/build-vm-kvm
+++ b/build-vm-kvm
@@ -173,6 +173,13 @@ vm_verify_options_kvm() {
             kvm_options="$kvm_options -device $item"
         done
     fi
+
+    if test -c /dev/hwrng ; then
+        rng_dev="/dev/hwrng"
+    else
+        rng_dev="/dev/random"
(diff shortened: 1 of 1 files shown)


andreas-schwab:
What happens if /dev/random is depleted?

k0da:
@andreas-schwab well, nothing I assume. My guess is, It wouldn't be worse.i

andreas-schwab:
It breaks all armv7 worker.

[   16s] qemu-system-arm: -device virtio-rng-pci,rng=rng0: No 'PCI' bus found for device 'virtio-rng-pci'

andreas-schwab:
If /dev/hwrng exists but has no drivers then qemu crashes.

ERROR:backends/rng-random.c:47:entropy_available: assertion failed: (len != -1)

PR #209, opened 2015-10-24
build-vm-kvm: fix check for presence of hw random device

andreas-schwab:
Even if /dev/hwrng exists there is no guarantee that a hw random device
exists.  If there is none it causes qemu to crash.

ERROR:backends/rng-random.c:47:entropy_available: assertion failed: (len != -1)


andreas-schwab:
This has nothing to do with arm

PR #210, opened 2015-10-26
[armv7l] Unbreak random number device pass through

dirkmueller:
The generic device name is virtio-rng-device which works for all
platforms. virtio-rng-pci is a specific implementation attached
tot he PCI bus. Unfortunately, ARMv7 does not have a PCI bus, so
everything fails with:

 No 'PCI' bus found for device 'virtio-rng-pci


adrianschroeter:
that is leading to other errors (assertion filaed len !=  -1)

dirkmueller:
@adrianschroeter well, we didn't have the hw random device driver compiled in yet.

PR #212, opened 2015-11-01
[build] add support for LXC 1.0 and 1.1 parameter calls

mmohring:
Current LXC support works only for LXC 1.1. But LXC 1.1 does not work with OBS, so check now for the version and use 1.0 or 1.1 calls.

### DIFF
--- a/build-vm-lxc
+++ b/build-vm-lxc
@@ -51,8 +51,16 @@ vm_startup_lxc() {
 	mount --bind "$BUILD_ROOT" "$LXCROOTFS"
 	EOF
     chmod a+x "$LXCHOOK"
-    lxc-create -n "$LXCID" -t none -f "$LXCCONF" || cleanup_and_exit 1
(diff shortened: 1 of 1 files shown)


adrianschroeter:
What are the problems with lxc 1.1 ?

mmohring:
Forgot to link issue #213 explaining the error messages of LXC 1.1. It explains it.

mmohring:
Is there any reason not to support LXC 1.1 *and* 1.0 ?

adrianschroeter:
I just wonder if there should be also support for 0.x . But let's merge and see. LXC environment is luckily not so critical as kvm/xen/chroot

mmohring:
No problem to add also calls for 0.x if they changed parameters compared to 1.0. Do we know anything about the errors lxc 1.1 is giving, e.g. issue #213 ? Is it a problem in the setup of the container when using 1.1 or is it a bug in 1.1 ?

PR #214, opened 2015-11-02
Add support for full author identities in changes files

Conan-Kudo:
This PR adds support for "Name <email>" format for author ID in changelog entries. This structure is common among virtually all package changelog formats now (both RPM %changelog and Debian changelog files use this structure).

### DIFF
--- a/vc
+++ b/vc
@@ -28,6 +28,14 @@ if [ -z "$mailaddr" ]; then
 	mailaddr="$USER@$domain"
 fi
 
+if [ -n "$VC_REALNAME" ]; then
+	packager="$VC_REALNAME <$VC_MAILADDR>"
+elif [ -x /usr/bin/rpmdev-packager ]; then
+	packager=`rpmdev-packager`
+else
+	packager="`getent passwd $UID | cut -d: -f5 | cut -d ',' -f 1` <$mailaddr>"
+fi
+
 EDITOR=${EDITOR:-vim}
 date=`LC_ALL=POSIX TZ=UTC date`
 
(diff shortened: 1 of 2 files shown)


adrianschroeter:
--newformat creates the impression that openSUSE wants to switch to that format, but this has not been discussed at all AFAIK.

Could you rename the option to something like "--with-name" or "--with-user-name"?

Also, it would be nice if the code would check if the former entry used this format and switches to this format automatically.

Conan-Kudo:
@adrianschroeter I've reworked this change as you asked (nearly a year ago!).

That said, I think it actually makes more sense to make this form the default or the expected form for changes files.

adrianschroeter:
Makes sense to me, but it would cause a change in our distributions.

Our .changes files could(!) have "Fullname <email>" instead of just "<email>" behind the - depending
on the packagers installation (having rpmdevtools installed or not).

@DimStar77 @lnussel @bugfinder any opinion if this is wanted, harmful or should be enforced?

adrianschroeter:
@mlschroe is there anything in rpm(-build) we need to take into account regarding encoding?
(eg. my name has an umlaut, so we won't have strict 7bit ascii anymore).

A test submission with such an entry is here: https://build.opensuse.org/request/show/567289

vc
```diff
@@ -28,6 +28,12 @@ if [ -z "$mailaddr" ]; then
 	mailaddr="$USER@$domain"
 fi
 
+if [ -x /usr/bin/rpmdev-packager ]; then
+	packager=`rpmdev-packager`
+else
+	packager="`getent passwd $UID | cut -d: -f5` <$mailaddr>"
```
lnussel: need to parse GECOS here

Conan-Kudo: Isn't that what I'm doing?

lnussel: no, you're just parsing passwd itself, not the GECOS field
 https://en.wikipedia.org/wiki/Gecos_field

Conan-Kudo: I think I got it now.

DimStar77:
Generally, I'm all for it, with a few concerns/things to validate

* factory-package-news: must not trip over this (invalid dates caused it to spit out completely changelogs in the past, something which only makes noise on the mailing lists)
* Other tools used that parse the changelog might trip over the changed format (need adjustment)

lnussel:
factory-package-news doesn't care about the author of changelog entries, just changelogtime and changelogtext

PR #218, opened 2015-11-12
build-vm: Resize VM images

eugenetan:
The VM creation was checking to see if a root image existed. In doing
so it skips subsequent changes to root image size.


gollub:
:+1:

PR #219, opened 2015-11-16
[build] for KVM builds make sure loop and dm modules are preloaded

mmohring:
These modules are needed for example by KIWI and may not be in the initrd on some build targets, for example non SUSE distributions.

### DIFF
--- a/build-vm-kvm
+++ b/build-vm-kvm
@@ -226,7 +226,7 @@ vm_startup_kvm() {
     set -- $qemu_bin -nodefaults -no-reboot -nographic -vga none $kvm_options \
 	-kernel $vm_kernel \
 	-initrd $vm_initrd \
-	-append "root=$qemu_rootdev $qemu_rootfstype $qemu_rootflags panic=1 quiet no-kvmclock nmi_watchdog=0 rw rd.driver.pre=binfmt_misc elevator=noop console=$kvm_console init=$vm_init_script" \
+	-append "root=$qemu_rootdev $qemu_rootfstype $qemu_rootflags panic=1 quiet no-kvmclock nmi_watchdog=0 rw rd.driver.pre=binfmt_misc rd.driver.pre=loop rd.driver.pre=dm-mod elevator=noop console=$kvm_console init=$vm_init_script" \
 	${VM_MEMSIZE:+-m $VM_MEMSIZE} \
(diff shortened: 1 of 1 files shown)


mmohring:
KIWI needs the loopback devices and dm module to create the image files when calling kpartx, losetup and other operations on mounted image files.

adrianschroeter:
hm,  I am not very keen to add this here. Can we be absolute sure that is not creating failures when eg. loop module or kernel support is missing? This may affect old builds then...

Can't this be solved in kernel-obs-build package? IMHO the guest OS should define the used initrd and kernel itself to avoid that we break other distro builds.

mlschroe:
rd.driver.pre is a dracut option, right? Doesn't that mean that the option only helps if the modules are in fact in the initrd?
If that's not the case, why is a normal modprobe insufficient?

mmohring:
It seems to work also on non dracut systems, build-vm-kvm passes currently "rd.driver.pre=binfmt_misc" . This was introduced when openSUSE was *not using dracut*. So I was under the impression that it is *not a dracut option*, because it is passed on build-vm-kvm unconditionally to qemu-kvm. Can it also be a kernel cmdline option ?

mmohring:
openSUSE 13.1 has no dracut. When building for OS 13.1, "rd.driver.pre=binfmt_misc" is passed to qemu-kvm. Dito for Ubuntu or Debian. So is it ignored there ?

mlschroe:
The rd.driver.pre option really makes only sense for the initrd, otherwise you simply can do a modprobe. And in that case it makes no sense to claim that openSUSE 13.1 has no drakut, as the initrd is taken from the worker machine (if you don't use a special obs kernel).

The commit that introduced rd.driver.pre is from March 2014.

mmohring:
So that means "rd.driver.pre=binfmt_misc" should also be removed, the arguments apply also for that case ?

mlschroe:
I don't know, to be honest. I don't remember why it was added.

bugfinder:
binfmt_misc was/is for the arm/aarch64 cross builds. as I think we're building natively for all these now, it can probably go.

adrianschroeter:
this is considered old code and should not be touched anymore. all new adaptions happen inside of the distributions in the obs initrd.

PR #221, opened 2015-11-27
allow skipping of unknown rpm specfile tags

seife:
Non-SUSE distributions (and maybe old SUSE distributions) do not know
RPM tags "Suggests", "Recommends", "Supplements".
This leads to ugly hacks needed in spec files if cross-distro build is
needed.
Test if these tags are actually supported by the target's RPM and just
drop them if not. This needs to be enabled explicitly in prjconf with
"Buildflags: dropunknownrpmtags"

Should be "mostly harmless" if not explicitly enabled in prjconf.

### DIFF
--- a/build-recipe-spec
+++ b/build-recipe-spec
@@ -85,6 +85,40 @@ recipe_prepare_spec() {
 	test -z "$ABUILD_TARGET" || echo "build target is $ABUILD_TARGET"
     fi
 
+    # if "Buildflags: dropunknownrpmtags is" set in prjconf, check if the target's rpmbuild can
+    # actually handle Suggests, Recommends and Supplements. If it can't, drop these lines.
+    DROP_UNKNOWN=$(queryconfig --dist "$BUILD_DIST" --configdir "$CONFIG_DIR" --archpath "$BUILD_ARCH" buildflags dropunknownrpmtags)
+    if [ -n "$DROP_UNKNOWN" ]; then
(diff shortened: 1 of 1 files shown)


lnussel:
well, if a setting in prjconf triggers that the check isn't needed, one could just configure what tags to remove. Also, substitutedeps might be the better place for the job as it mangles the spec file already.

seife:
I was actually thinking about configuring what tags to remove.
But then I found that the automatic detection works just fine (and only removes the tags if it's really necessary -- once newer RPM hits RedHat/Mandriva, it wil just turn into a noop).

I'll look into substitutedeps if this is a better place for this specfile mangling.

adrianschroeter:
This is too specific code for the build script in our opinion. But you should be able to turn this into a source service and enable it at buildtime. That way you can configure it also per repo and you can also test if the installed rpm is supporting the tags.

seife:
But a source service will change the spec file for all distributions on checkin, wouldn't it?.
The "elegance" of my approach is, that I can use an unmodified spec file which builds for SUSE distributions, just add a RHEL target and it will "Just work™" :-)

But anyway, I can just keep my trivial patch in my build service instances.

adrianschroeter:
The source service with mode="buildtime" would run inside of the build environment. 

So, your spec file in your soruce repo would stay untouched, it just gets modified when needed right before rpmbuild is called inside of the build environment.

seife:
Thanks, I did not know that, will investigate. In this case, rest in peace, patch :-)

PR #236, opened 2015-12-18
Arch Pacman with auto-mode can't build i686 binary in x86_64 host

illuusio:
Arch Pacman with auto-mode can't build i686 binary in x86_64 host. This one forces Pacman to i686 or x86_64 mode and so it can't be too smart with build.


illuusio:
This cabe closed

PR #237, opened 2016-01-21
PowerVM backend

k0da:


adrianschroeter:
e75dcbd might create a security issue, because data from former builds will be still available. Needs to be double checked how this works for all LVM, plain block, loop file,.... situations.

I am fine to take the PowerVM backend patch.

adrianschroeter:
k, after a second look, your code is fine, we need to solve the re-allocation elsewhere correctly :)

PR #239, opened 2016-02-01
Reenable monitor mode for serial (stdin/out)

dirkmueller:
The monitor mode is useful to debug hung build hosts and shouldn't
be a security concern as it requires access to the screen(1) of the
build worker, and when you have that you can do nasty things anyway.


adrianschroeter:
we want to have this only via an extra switch. but we need to bring this in sync with the other debug options and also supporting it via osc. needs a second thought ....

adrianschroeter:
implemented  now with a switch in build, bs_worker and obsworker init script.

Issue #240, opened 2016-02-06
debtransform considers files ending with .tar as source files. 


hurricanehrndz:
files ending in .tar should not be considered source otherwise you get:
```
 Too many files looking like a usable source tarball (would not know which to pick): termite-v10.32.tar.gz termite-v10.32.tar
```

dkgroot:
I notice the same thing. Further investigation on my part shows that the service:recompress progress does not remove the original tar file after it has compressed it to one of: gz, xz, bz2. That causes debtransform to see two files which in turn then failes with the error mentioned above.

hurricanehrndz:
@dkgroot 

I have created an issue with the service:recompress. Unfortunately after analyzing the code I could not find fault with the bash script. [Line 118 of recompress](https://github.com/openSUSE/obs-service-recompress/blob/master/recompress) is suppose to remove the original tar file. I'm not sure why that does not occur when obs-build calls the service. Maybe core.py from obs-build needs to get analyze as to why the file is prevent from being removed. Either which way [Debian source package format](https://wiki.debian.org/Projects/DebSrc3.0) does not support tar compression alone and thus should be ignored. Hopefully this pull request gets accepted.

dkgroot:
Skipping .tar as a possible source file makes sense in any case. Agreed. 
Investigating what changed in recompress can be handled as a seperate issue. I do remember the original tar file being removed after recompress before. Maybe even as early as two days ago.

hurricanehrndz:
@dkgroot 

Me too, 3 hours.

dkgroot:
;-). Keeping fingers crossed !

dg0yt:
For those relying on tar_scm+recompress+set_version, and looking for a workaround:

Use an additional download_url service to fetch a tarball with a fixed name before tar_scm. The tarball's name can be used in the DSC for the tag  ```DEBTRANSFORM-TAR```. tar_scm is still used for set_version if it comes after download_url.
This adds another tarball to fix the problem from having two...

dkgroot:
@dg0yt : Thanks i already figured that one out, problem is our repository (on sourceforge) does not provide regular tarball download from trunk without some kind of user intervention. But thanks for mentioning it anyway, might be helpfull for someone else :-)

dg0yt:
I would prefer if tar_scm would create compressed tarballs. Submitted untested attempt at https://github.com/openSUSE/obs-service-tar_scm/pull/93.

dkgroot:
[...]

PR #241, opened 2016-02-06
Fixing issue:https://github.com/openSUSE/obs-build/issues/240

hurricanehrndz:
Sometime recompress fails to remove the original tar file. So debtransform should ignore all files ending in tar as there is no debian source format that supports files with only tar compression as source files.


dg0yt:
Just dropping all .tar from the list isn't the right approach IMHO. There might be reasons for different compressed archives to exist. DEBTRANSFORM-TAR should accept shell patterns to allow for disambiguation even when the name is not known exactly (tar_scm).

adrianschroeter:
it was a temporary bug in the OBS service that the .tar files got not removed. This is fixed, so this should not be needed anymore.

PR #244, opened 2016-02-15
aggregate patch fields in spec file

M0ses:


M0ses:
In preparation of

https://github.com/openSUSE/obs-service-download_files/issues/6

PR #249, opened 2016-03-13
pass SOURCE_DATE_EPOCH into build env

bmwiedemann:
to allow reproducible builds

for this to work from osc build, you need to add SOURCE_DATE_EPOCH
to the Defaults env_keep line in /etc/sudoers

### DIFF
--- a/build
+++ b/build
@@ -411,6 +411,8 @@ shellquote() {
 # through /bin/su -c
 toshellscript() {
     echo "#!/bin/sh -x"
+    SOURCE_DATE_EPOCH=${SOURCE_DATE_EPOCH//[^0-9]/} # sanitize
+    echo "export SOURCE_DATE_EPOCH=$SOURCE_DATE_EPOCH"
     echo -n exec
     shellquote "$@"
     echo


adrianschroeter:
editing sudoers on all workers is not really a good idea.

Doesn't it make sense to set it based on the .changes file entry maybe?
So we would not need to hand it over inside the VM.
Or maybe a buildtime source service could do this?

bmwiedemann:
yes, could be done like in https://build.opensuse.org/package/view_file/games/falconseye/falconseye.spec
```bash
if test -z "$SOURCE_DATE_EPOCH" ; then
  modified="$(sed -n '/^----/n;s/ - .*$//;p;q' "%{SOURCE99}")"
  SOURCE_DATE_EPOCH=$(date -d "${modified}" "+%%s")
fi
```

but we would need to have the .changes file available in every .src.rpm to allow to rebuild the same binary rpm.

adrianschroeter:
we could just use the latest mtime of all source files then maybe?

Why is the src.rpm important here?

adrianschroeter:
closing for now

PR #253, opened 2016-03-22
Allow AArch32 builds on AArch64 hosts

agraf:
Most AArch64 systems are able to run KVM VMs with AArch32 kernels. Allow
that combination and pass the correct parameters into qemu in that case.

To make sure we don't break armv8l builds that would get executed on an
AArch64 kernel, only map armv5/6/7.

Signed-off-by: Alexander Graf <agraf@suse.de>

### DIFF
--- a/build-vm
+++ b/build-vm
@@ -750,7 +750,7 @@ vm_first_stage() {
     test "$PERSONALITY" = -1 && PERSONALITY=0	# syscall failed?
     case $(uname -m) in
 	ppc|ppcle|s390) PERSONALITY=8 ;;	# ppc/s390 kernel never tells us if a 32bit personality is active, assume we run on 64bit
-	aarch64) test "$BUILD_ARCH" != "${BUILD_ARCH#armv}" && PERSONALITY=8 ;; # workaround, to be removed
+	aarch64) [[ "$BUILD_ARCH" == armv[567]* ]] && PERSONALITY=8 ;; # workaround, to be removed
     esac
     test "$VM_TYPE" = lxc -o "$VM_TYPE" = docker && PERSONALITY=0
     echo "PERSONALITY='$PERSONALITY'" >> $BUILD_ROOT/.build/build.data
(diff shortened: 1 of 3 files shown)


common_functions
```diff
@@ -29,7 +29,9 @@ build_host_arch() {
 
 extend_build_arch() {
     case $BUILD_ARCH in
-      aarch64) BUILD_ARCH="aarch64:armv8l" ;;
+      aarch64) BUILD_ARCH="aarch64:armv8l"
+	       [ "$VM_TYPE" = kvm ] && BUILD_ARCH="$BUILD_ARCH:armv7hl:armv7l:armv6hl:armv6l:armv5tel"
+	       ;;
```
adrianschroeter: Why is this needed? armv8l is really working in plain chroot on aarch64?

build-vm
```diff
@@ -750,7 +750,7 @@ vm_first_stage() {
     test "$PERSONALITY" = -1 && PERSONALITY=0	# syscall failed?
     case $(uname -m) in
 	ppc|ppcle|s390) PERSONALITY=8 ;;	# ppc/s390 kernel never tells us if a 32bit personality is active, assume we run on 64bit
-	aarch64) test "$BUILD_ARCH" != "${BUILD_ARCH#armv}" && PERSONALITY=8 ;; # workaround, to be removed
+	aarch64) [[ "$BUILD_ARCH" == armv[567]* ]] && PERSONALITY=8 ;; # workaround, to be removed
```
adrianschroeter: I believe this is not working with bash v3. Can we keep the test command please?

common_functions
agraf: Yes, all you need to do is run linux32 and you have an armv8l system.

build-vm
agraf: So how would the test version look? We'd have to do 3 checks here. The [[ ]] checks have been in bash forever - I don't see a problem here.

PR #254, opened 2016-03-22
Allow AArch32 builds on AArch64 hosts

agraf:
Most AArch64 systems are able to run KVM VMs with AArch32 kernels. Allow
that combination and pass the correct parameters into qemu in that case.

To make sure we don't break armv8l builds that would get executed on an
AArch64 kernel, only map armv5/6/7.

Signed-off-by: Alexander Graf <agraf@suse.de>

### DIFF
--- a/common_functions
+++ b/common_functions
@@ -29,7 +29,11 @@ build_host_arch() {
 
 extend_build_arch() {
     case $BUILD_ARCH in
-      aarch64|aarch64_ilp32) BUILD_ARCH="aarch64:aarch64_ilp32:armv8l" ;;
+      aarch64|aarch64_ilp32)
+	BUILD_ARCH="aarch64:aarch64_ilp32:armv8l"
+	# With KVM we can also run AArch32 builds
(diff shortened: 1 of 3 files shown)


agraf:
Ping?

common_functions
```diff
@@ -29,7 +29,11 @@ build_host_arch() {
 
 extend_build_arch() {
     case $BUILD_ARCH in
-      aarch64|aarch64_ilp32) BUILD_ARCH="aarch64:aarch64_ilp32:armv8l" ;;
+      aarch64|aarch64_ilp32)
+	BUILD_ARCH="aarch64:aarch64_ilp32:armv8l"
+	# With KVM we can also run AArch32 builds
+	[ "$VM_TYPE" = kvm ] && BUILD_ARCH="$BUILD_ARCH:armv7hl:armv7l:armv6hl:armv6l:armv5tel"
```
dirkmueller: this needs an extra flag..

adrianschroeter:
merged with a minor change.

PR #257, opened 2016-03-31
[build]: added DoD for aarch64 repositories on Debian

mmohring:
Added a mapping for aarch64 architecture which is called arm64 in the Debian meta data.


mmohring:
With this change for example Debian 8 on aarch64 works in a DoD descriptor:

`<download arch='aarch64' url='http://ftp.de.debian.org/debian/jessie/main' repotype='deb'>`

PR #265, opened 2016-05-27
fix macro expansion for macros without curly braces

StefanBruens:
Parametrized macros without curly braces consume all remaining arguments
until end-of-line.

Fixes evaluation of e.g "%define foo() %1", which caused "Use of
initialized value $line in ..." due to dereference of uncaputured $2.


mlschroe:
Thanks!

PR #269, opened 2016-06-09
Fix dpkg force all

jblunck:
Since dpkg 1.18.5 (Debian 9) --force-all implied --force-script-chrootless which will execute the maintainer scripts outside of the chroot even if they don't support that.

This fixes the installation of the chroot on Debian 9.

### DIFF
--- a/build-pkg-deb
+++ b/build-pkg-deb
@@ -21,6 +21,25 @@
 #
 ################################################################
 
+#
+# A wrapper around chroot to set the environment correctly for dpkg and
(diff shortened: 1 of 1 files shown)


mlschroe:
Sorry, those directory changes are not pointless. They are a workaround for some version of fileutils that doesn't do a 'chdir /' when doing a 'chroot /' call.
Please remove that commit from the pull request.

jblunck:
I've pulled the chdir into the deb_chroot helper and added a comment.

PR #275, opened 2016-06-22
Add architecture-specific sources extract for Arch Linux

AlexTalker:
Since PKGBUILD support architecture-specific sources, this is required for properly support some of them. Only 2 architectures supported officially.

### DIFF
--- a/Build/Arch.pm
+++ b/Build/Arch.pm
@@ -91,6 +91,9 @@ sub parse {
   push @{$ret->{'deps'}}, @{$vars{'checkdepends'} || []};
   push @{$ret->{'deps'}}, @{$vars{'depends'} || []};
   $ret->{'source'} = $vars{'source'} if $vars{'source'};
+  # Maintain architecture-specific sources for officially supported architectures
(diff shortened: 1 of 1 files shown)


AlexTalker:
Here's a testing with a bit of Dumper on STDERR.
https://gist.github.com/7fd27be0f8c59e6fb8b138480ef60f30
All data used by default is only PKGBUILD. I did test it only using download_file service and build since develop properly environment for whole thing might be too long with my slow internet connection. I hope that you will soon merge this PR and roll on build.opensuse.org. Thank you for the amazing useful build service!

AlexTalker:
Here's an example of same PKGBUILD(except verison and hashes, I guess) which doesn't build on current obs https://build.opensuse.org/package/show/home:alextalker:aur/firefox-aurora

AlexTalker:
Can you pull this fix, please?

mlschroe:
Thanks!

AlexTalker:
@mlschroe thank you too! :smile: Could you please notify me when the fix will be rolled on build.opensuse.org?

PR #278, opened 2016-07-04
[ArchLinux] Set up the default packager information

ItachiSan:
Default is the OpenSUSE Open Build Server.
I wrote this unaware if the packager information are available somewhere as variable.


adrianschroeter:
Correct name would be "openSUSE Build Service"

adrianschroeter:
we should not hardcode the packager in the code here. In rpm case we use the %packager macro, which can be defined in OBS project configs. We should do something similar in ARCH case.

ItachiSan:
@adrianschroeter just to know, this `packager` information is available as... environment variable?
Because I can probably set the value properly then in the proper script.

PR #279, opened 2016-07-09
Don't forget of architecture-specific dependencies

ItachiSan:
Right now, the build system ignores them and this is not really nice.
This should do the trick (not sure, I'm learning Perl fixing stuff here almost).


mlschroe:
(I changed the code to only add deps of the matching architecture.)

ItachiSan:
@mlschroe you did well, I don't know Perl so thanks a lot :D

PR #280, opened 2016-07-09
Use source information properly

ItachiSan:
All information needed are present in the comment in the code.

Obviously, feel free to ask if something is not clear :smile:

### DIFF
--- a/Build/Arch.pm
+++ b/Build/Arch.pm
@@ -203,9 +203,11 @@ sub query {
     s/:.*// for @suggests;
     $ret->{'suggests'} = \@suggests if @suggests;
   }
-  # arch packages don't seem to have a source :(
-  # fake it so that the package isn't confused with a src package
-  $ret->{'source'} = $ret->{'name'} if defined $ret->{'name'};
+  # Keep consistent to Arch style; refer to sources as the upstream sources
+  # The 'source' array is always defined and referred to sources needed for building the package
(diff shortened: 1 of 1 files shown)


Build/Arch.pm
```diff
@@ -203,9 +203,11 @@ sub query {
...
     $ret->{'suggests'} = \@suggests if @suggests;
   }
-  # arch packages don't seem to have a source :(
-  # fake it so that the package isn't confused with a src package
-  $ret->{'source'} = $ret->{'name'} if defined $ret->{'name'};
+  # Keep consistent to Arch style; refer to sources as the upstream sources
+  # The 'source' array is always defined and referred to sources needed for building the package
+  $ret->{'source'} = $ret->{'source'} || [];
+  push @{$ret->{'source'}}, @{$ret{'source_i686'} || []};
+  push @{$ret->{'source'}}, @{$vars{'source_x86_64'} || []};
```
adrianschroeter: $vars should be $ret ?

mlschroe:
The current code can't deal with $ret->{'source'} being an array. It's not important for OBS, anyway. It's pretty much just used as a marker that this is not a source package.

ItachiSan:
I still think is should be nice to have such an information.
Also, from what I've understood, the 'source' variable holds the upstream sources and it could also download them, as `makepkg` does (a feature currently unavailable on OBS).

PR #283, opened 2016-08-01
Consider the "Icon" tag in Rpm::parse

marcus-h:
This is needed by https://github.com/openSUSE/obs-service-source_validator/pull/39


marcus-h:
I just updated the PR/commit - new description:
Consider the Icon tag and fixed detection of Source, Patch and Url tags

Source, Patch and Url tags can also appear in the preamble of a
subpackage.

PR #289, opened 2016-09-06
Add optional whitelisting for allowed buildroots and parameters

dwmh:
Co-Author: Juha Kallioinen <juha.kallioinen@ericsson.com>

Can be used to make multi-user environments more secure. Does not change
functionality if /etc/obs/build.conf is not present on system.

### DIFF
--- a/build
+++ b/build
@@ -95,6 +95,10 @@ CCACHE=
 DLNOSIGNATURE=
 CACHE_DIR=/var/cache/build
 
+# Input validation
+BUILDROOT_VALID=false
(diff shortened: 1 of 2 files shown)


mlschroe:
Thanks for the patch! I'm rewriting it a bit and will commit it next Monday.

mlschroe:
Ok, pushed. I rewrote it quite a bit, though.

Oh, and I use /etc/build.conf as configuration file, as this tool has nothing to do with OBS, it's just used in OBS for building.

PR #290, opened 2016-09-08
Prevent wide character warning

lethliel:
@mlschroe please review

switching the mode of the filehandle to utf8 gets rid of the "wide characters in print" warnings.


lethliel:
Offline review with mlschroe. Can be merged.

PR #292, opened 2016-09-12
build-vm-kvm: use virtio-rng-device also on aarch64

andreas-schwab:
This is needed so that it works with an armv7 guest.

### DIFF
--- a/build-vm-kvm
+++ b/build-vm-kvm
@@ -108,6 +108,7 @@ vm_verify_options_kvm() {
 		kvm_options+="-M virt"
 	    fi
 	    kvm_device=virtio-blk-device
+	    kvm_rng_device=virtio-rng-device
 	    ;;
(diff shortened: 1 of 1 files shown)


dirkmueller:
Hmm, maybe we only want to do that in the armv8l case? I am not sure we want to use mmio for aarch64 native

dirkmueller:
@agraf? I'm okay with giving it a try, maybe we can deploy it in obs and see how it works out?

andreas-schwab:
We already use mmio for the virtio-blk device anyway.

agraf:
The breakage happens when building armv7l code on aarch64 hosts. It also shouldn't break anything on aarch64 guests, so yes, this change is good.

adrianschroeter:
it is deployed now for a bit more testing. will merge this the next days if no problems appear.

PR #295, opened 2016-09-21
Add option to wipe a buildroot

dwmh:
Build creates root owned files, so this option allows build to be used to also delete those files.

With this, normal users can delete their build-roots while only having sudo rights to the build script.


dwmh:
Corresponding osc pull request:
https://github.com/openSUSE/osc/pull/230

dwmh:
Added rc check for the umounts.

PR #296, opened 2016-09-28
configs/debian: add linux-libc-dev to Keep list

bluca:
The Debian/Ubuntu kernel source package builds some userland tools
(perf, etc), so it needs linux-libc-dev. But since it is built from
the same source package, the dependency resolver blacklists it,
making it impossible to build Debian/Ubuntu kernel packages.
Add Keep: linux-libc-dev to the default config to avoid the
blacklisting.

Fixes: https://github.com/openSUSE/obs-build/issues/300


jblunck:
Looks good to me.

PR #299, opened 2016-10-11
Retrieve exitcode from _exitcode for LXC/libvirt

luizluca:
virsh does not return the error code (other than its own errors)
nor build can retrieve it from swap. So, every failed build was
passed as success.

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>


mlschroe:
Thanks!

PR #303, opened 2016-10-28
Recognize Recommends relations of RPMs.

### LINKED ISSUE #302: Cannot handle "recommends" properly.
Case:

   Package A
      A requires A-compat
      A recommends A-x
         A-x provides A-compat
         A-y provides A-compat

   Package B
      B buildrequired A

Result of building B

 "unresolvable" because "have choice for A-compat needed by A: A-x A-y

Expected Result:

  Building B incurs installing A-x and A unless it explicitly BuildRuquires on A-y
## 

Suggestion:

   Handle "recommends" in obs-build and resolve it at Build.pm's expand function.

Workaround being tested:

  Right before "expand" emits "have choice for", if additional OBS project config says "resolve_choice=alphabetically" or so, it simply choose one of the candidates.

myungjoo:
When there are multiple candidate packages as a result of
Provides, the current obs-build gives an error: "have choice for".

However, according to http://www.rpm.org/wiki/PackagerDocs/Dependencies ,
Recommends tells the system to install the corresponding package
unless there is a conflict; thus, if there is a package recommended
(as described in issue #302), obs-build shall install the recommended
package, resolving "have choice for" error.

Fixes #302

Signed-off-by: MyungJoo Ham myungjoo.ham@samsung.com

### DIFF
--- a/Build.pm
+++ b/Build.pm
@@ -678,8 +678,8 @@ sub get_cbinstalls { return (); }
 
 sub readdeps {
   my ($config, $pkginfo, @depfiles) = @_;
-
   my %requires;
+  my %recommends;
   local *F;
   my %provides;
   my %pkgconflicts;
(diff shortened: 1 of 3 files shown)


myungjoo:
In this commit, I've added a new shortened tag "r", which represents "Recommends". I've used small r because capital R is already taken by "Requires" and I couldn't find a document stating standard lettering of package dependencies.

Conan-Kudo:
@myungjoo It should also recognize `Suggests` for this, as this matches behavior with libsolv.

myungjoo:
@adrianschroeter Like the pull-request for perl-BSSolv, this commit is also updated to be enabled optionally with "BuildFlags: UseRecommendsForChoices"

PR #307, opened 2016-11-15
livebuild: parse /etc/debian_version on testing

bluca:
When ran on Debian testing/unstable, lsb_release --codename will
return "n/a", which results in a broken config.
As a workaround, if "n/a" is returned, parse /etc/debian_version
which will contain "\<testing codename\>/sid".

### DIFF
--- a/build-recipe-livebuild
+++ b/build-recipe-livebuild
@@ -122,6 +122,15 @@ recipe_build_livebuild() {
 
     [ -z "${ARCH}" -o -z "${DIST}" ] && cleanup_and_exit 1
 
+    # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=845651
+    # For our use case it makes sense to always return the testing codename,
+    # as creating a dependency on live-build-desc-sid would not make sense. In
+    # the example cited in the above bug, the metadata for sid would be incorrect
+    # anyway, and we would want the ones for potato.
+    if test "${DIST}" = "n/a" ; then
+	DIST=$(chroot $BUILD_ROOT su -c "sed 's/\(.*\)\/.*/\1/' /etc/debian_version")
+    fi
+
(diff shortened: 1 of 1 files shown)


adrianschroeter:
@jblunck can you comment on this please?

jblunck:
If I understand it correctly lsb_release is already looking at /etc/debian_release. Maybe its better to apply a fix to lsb_release if that isn't working properly.

bluca:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=845651

adrianschroeter:
so it is supposed to be fixed on debian side and we can skip this?

bluca:
It can be fixed there, doesn't mean it will :-)
I will close for now and reopen if there's a negative answer (or no answer for too long) from the lsb-release maintainer

bluca:
The maintainer of lsb-release correctly pointed out that if apt fails there's no way to differentiate between testing and unstable, so returning n/a is correct in the sense that in the context of lsb-release there's no way to be sure.

One could argue that for OBS only the Debian testing repository is provided so the answer is always going to be testing, but on a private instance a user could in principle set up an unstable build.

So this is mostly for the benefit of someone coming here through google looking for a resolution, as I think it's right not to fix it this way.

bluca:
Reopening as asked by @jblunck

jblunck:
I believe for our purpose it is ok to return the testing codename (not "sid") in case we build against unstable. If we would build today (testing is "stretch") against an old "unstable" repository with 'potato/sid' it wouldn't make sense to create a dependency to the "live-build-desc-sid" package. It is very likely that this package would contain metadata for the "sid" distribution that is going to become "stretch".

@bluca It would help for future understanding if you add a comment explaining why we choose to not return 'n/a' in this case that also reference the discussion in the lsb-release bug.

bluca:
@jblunck done

jblunck:
Looks good to me.

PR #308, opened 2016-11-17
Compatibility fixes for newer Linux Kernel versions related to mounts in chroots.

Ionic:
See the per-commit descriptions.

This fixes the issue of dangling mount when using `obs-build` with chroots on systems with a relatively recent Linux Kernel version.


adrianschroeter:
commit 5dca4519767175e37421ae7e0247d577fbbc578e should have solved this. Can you please verify that it works for you?

mlschroe:
(about the /dev/pts/ptmx mode: you aren't supposed to access that device node at all, you should open /dev/ptmx with has mode 666)

PR #311, opened 2016-12-01
build-recipe-kiwi: export RELEASE to kiwi build

seife:
Make RELEASE variable available to the KIWI build environment.
This can be used in build.sh/images.sh like this:
  cp -a /image/.profile /kiwi-deploy/profile
  echo "kiwi_obsrelease=Build$RELEASE" >> /kiwi-deploy/profile
and made available to the installed image.

Of course a much better solution would be:
* add a KIWI option --release
* use that KIWI option to obs-build.
But this is a low-tech, yet working alternative which has no other dependencies.
If deemed useful, I can try to work with kiwi upstream to implement that as a long-term solution.

### DIFF
--- a/build-recipe-kiwi
+++ b/build-recipe-kiwi
@@ -409,7 +409,7 @@ perform_image_build() {
 	echo "running kiwi --prepare for $imgtype..."
     fi
     # Do not use $BUILD_USER here, since we always need root permissions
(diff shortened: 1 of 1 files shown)


adrianschroeter:
the --release kiwi option is something what we agreed since a longer time with kiwi people IIRC. Will check back if we can get this soonish ...

adrianschroeter:
this is solved meanwhile via --build-id parameter for kiwi

PR #318, opened 2017-01-04
Copy qemu linux user to chroot

kukoarmas:
When cross-compiling with a qemu-linux-user supported arch, the init_buildsystem script fails as soon as we chroot, because we registered de arch binformat to the qemu binaries in /usr/bin, and as soon as we chroot they are not there anymore

This simple fix just copies the needed qemu binaries for the arch we are building into $BUILD_ROOT/usr/bin

### DIFF
--- a/init_buildsystem
+++ b/init_buildsystem
@@ -588,10 +588,14 @@ else
 	    if test "$mmap_min_addr" != 0 ; then
 		echo "Warning: mmap_min_addr is != 0. If programs fail at mmap this could be the reason."
 	    fi
-	    # FIXME: be more clever on which qemu binaries we need
 	    echo "copying qemu-linux-user binaries to $BUILD_ROOT/usr/bin"
 	    mkdir -p $BUILD_ROOT/usr/bin
-	    cp -av /usr/bin/qemu-* $BUILD_ROOT/usr/bin
+	    for arch in `echo $BUILD_ARCH | tr ":" " "`; do
+		if [ -x /usr/bin/qemu-$arch-binfmt ]; then
+		    echo "  Copying qemu-$arch*"
+		    cp -av /usr/bin/qemu-$arch* $BUILD_ROOT/usr/bin
+		fi
+	    done
 	fi
     fi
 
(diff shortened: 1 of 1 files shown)


adrianschroeter:
The build depends on the version of qemu, so the qemu binaries need to be part of the build configuration (and get preinstalled. Check openSUSE:Factory:ARM project on build.opensuse.org for example).

We could offer a copy in optionaly, but only after a warning to the user, so that he has to use some switch which makes it clear that this is no reproducable build anymore.

kukoarmas:
Maybe I'm doing something incredibly stupid, or I didn't understand how it should work. Maybe I need to specify some option (--vm-type or something), but I've tried several options and none of them worked to build locally

I have copied the buildconfig from the openSUSE:Factory:ARM and tried to build without this patch and also failed as soon as a chroot is done:

[   13s] running aaa_base preinstall script
[   13s] chroot: failed to run command 'sh': No such file or directory
[   13s] running aaa_base postinstall script
[   13s] chroot: failed to run command 'sh': No such file or directory
[   13s] initializing rpm db...
[   13s] chroot: failed to run command '/usr/bin/rpmdb': No such file or directory

The problem again is that I don't have the /usr/bin/qemu-aarch64* x86 binaries inside the chroot

There are two very closely dependent actions:
- Registering the aarch64 binary format to be run using the /usr/bin/qemu-aarch64
- Having a valid /usr/bin/qemu-aarch64 for the hostarch (in my case, a x86_64 binary)

Since we setup the first action in the init_buildsystem script, I guessed that was the right place to make sure the correct binary was in the BUILD_ROOT before any chroot is done

Regarding the requirement to install qemu-linux-user, its defined in the buildconfig, but it's not enough, we need to have the binary inside the BUILD_ROOT before any chroot is done

The build is reproducible as long as the host system has the qemu-linux-user package installed

Maybe this is not the cleanest way, as wight now I don't even warn the user if the required qemu binary is not found, maybe I can add that warning

BTW, I also needed to patch build.py from osc to add "aarch64" as an arch that can be build in a x86_64 host...

PR #325, opened 2017-01-22
use fixed hash seed values for perl+python

bmwiedemann:
to allow packages like
perl-Apache-SessionX and python3-jupyter_nbconvert-doc
that output unordered hash values into the rpm
to have reproducible results (to make build-compare happy)
without finding and patching each such package individually

### DIFF
--- a/build
+++ b/build
@@ -438,6 +438,8 @@ shellquote() {
 # through /bin/su -c
 toshellscript() {
     echo "#!/bin/sh -x"
+    echo "export PYTHONHASHSEED=42"
+    echo "export PERL_HASH_SEED=42"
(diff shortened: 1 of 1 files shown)


bmwiedemann:
Holger Levsen (working on reproducible builds for Debian) proposed to set them to $SOURCE_DATE_EPOCH instead to provide some pseudo-randomness, but IMHO that is not really needed.

adrianschroeter:
Hm, can't we do this in some packages because it belongs to the content of the build environment?

bmwiedemann:
will try to get it into post-build-checks rpm

PR #336, opened 2017-02-08
new openstack build version

M0ses:


M0ses:
@adrianschroeter @mlschroe 

Please review

adrianschroeter:
How do we deal with the new dependency of python for os-console? 
Should it be a Recommends in spec file?

Are you sure you want to maintain a fallback code which would need extra testing?
I suppose you used some python lib which isn't easily available in perl, right?

mlschroe:
All the nova tools also need to be there (and they are also written in python). No recommends, please. But there should be a nice error message if 'nova' is not available...

mlschroe:
(The license of the tool is a bigger issue, as it is the apache license)

M0ses:
@mlschroe 

* I added some code to check for nova/cinder - hope you like it :-)
* about Apache License I have no clue how this could hit us. I could also write a console in perl - if it is required - or I could ask @larsks for permission to reuse his code.

larsks:
Hi there! You are absolutely welcome to re-use my code.  If you make improvements it would be great if you can submit them as prs, but I'm not really picky.  I just wrote that code because I couldn't find anything else to perform the task, and I'm happy if people find it useful.

M0ses:
@adrianschroeter I did not got your point. 

* What do you mean with "fallback code" ?
* I think the used modules in os-console/client are more common in python the perl equivalents

M0ses:
@larsks 

Thanks for quick response and great to hear that we could reuse your work.

I doubt, that my modifications on the code are useful for anybody than us, so I made no PR

BTW:
Thanks for your work on novaconsole - it helped me a lot and I hope novaconsole will get part of the official openstack distribution.

PR #342, opened 2017-03-13
dsc: Keep everything

sjoerd-ccu:
Debian package build dependencies allow for direct or indirect build
cycles, when doing a full distribution build in OBS this means keeping
the keep list gets very tedious very quickly.

Adjust the default keep processing such that for Debian (dsc) builds
everything is kept by default

--

This feels like a bit of hack and changes how OBS looks at dsc build type quite a bit. Another option would be to introduce a new configuration item to enable this mode specifically.

### DIFF
--- a/Build.pm
+++ b/Build.pm
@@ -537,7 +537,9 @@ sub get_build {
     @directdepsend = grep {!/^-/} splice(@directdepsend, @deps + 1);
   }
   my @extra = (@{$config->{'required'}}, @{$config->{'support'}});
-  if (@{$config->{'keep'} || []}) {
+  if ($config->{'type'} eq 'dsc') {
+	  ;
+  } elsif (@{$config->{'keep'} || []}) {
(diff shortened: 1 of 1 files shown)


mlschroe:
But isn't it more sane to just delete all 'Keep' lines from the debian project config?

sjoerd-ccu:
@mlschroe Afaict remove all the keep lines doesn't change the behaviour to keep everything ? Is this the intention (if so i can look at why it didn't work for us)?

My memory might be wrong, the patches where a while ago but i'm pretty sure i tried that..

alee-ccu:
Hi, I am `obs-build` package maintainer in Debian:
We currently carry this patch in our debian package:
    https://sources.debian.org/src/obs-build/20180302-3/debian/patches/0005-dsc-keep-everything.patch/

It would be nice if we may included this patch in upstream.

PR #343, opened 2017-03-13
debootstrap: generate apt caches

sjoerd-ccu:
Some packages seemingly rely on apt-cache (with populated caches) being
available during build. This seems an ugly assumption, but it does work on
the Debian buildd infrastructure so make it work on OBS as well.


adrianschroeter:
hm, can we really be sure that apt-cache executable is always there? Or should this maybe checked before?

nickbroon:
Yes, apt-cache should always be there. The chroot is generated with 'debootstrap --variant=buildd' which will install it. Even the 'minbase' variant will ensure that 'apt' package, that contains apt-cache, is installed.

PR #345, opened 2017-03-13
deb: Don't create an empty rcS

sjoerd-ccu:
rcS is a configuration file created by initscripts, not entirely clear
why obs-build tries to create it. But if it does initscripts will ask
whether to keep/replace it interactively blocking the build.

### DIFF
--- a/build-pkg-deb
+++ b/build-pkg-deb
@@ -48,7 +48,6 @@ deb_setup() {
     :>> $BUILD_ROOT/var/lib/dpkg/available
     :>> $BUILD_ROOT/var/log/dpkg.log
     :>> $BUILD_ROOT/etc/ld.so.conf
-    :>> $BUILD_ROOT/etc/default/rcS
 }
 
 pkg_initdb_deb() {


mlschroe:
Creating an empty rcS file was added in 2007 to make building work for some old debian releases. We can try to delete it again now in the hope it doesn't break any current builds.

dbnicholson:
If I had to guess, I'd say that this was a bug in old versions of the `initscripts` package. Digging through [the changelog from debian wheezy](http://metadata.ftp-master.debian.org/changelogs/main/s/sysvinit/sysvinit_2.88dsf-41+deb7u1_changelog), this was likely fixed in 2 places:

sysvinit (2.86.ds1-64):
  * Make sourcing of /etc/default/rcS conditional in init.d/rc, to
    get rid of lintian error.

sysvinit (2.88dsf-13.3):
  * debian/src/initscripts/lib/init/vars.sh:
  Read /etc/default/rcS and provide defaults if unset.

Looking at `/etc/init.d/rc` and `/lib/init/vars.sh` on wheezy, both seem to deal fine without `/etc/default/rcS` being present. So, I think that unless someone is trying to build a debian release from before wheezy (debian's oldoldstable), then this is likely not an issue.

PR #359, opened 2017-05-30
Run post-build-check scripts at the same run-level in parallel.

peti:
Run multiple post-build scripts in parallel if they have the same run-level.

Related PR to the post-build-checks: https://github.com/openSUSE/post-build-checks/pull/10

### DIFF
--- a/build
+++ b/build
@@ -1439,9 +1439,19 @@ if test -n "$RPMS" -a -d "$BUILD_ROOT/usr/lib/build/checks" ; then
     for SRPM in $BUILD_ROOT/$TOPDIR/SRPMS/*src.rpm ; do
 	test -f "$SRPM" && PNAME=`rpm --nodigest --nosignature -qp --qf "%{NAME}" $SRPM`
     done
-    for CHECKSCRIPT in $BUILD_ROOT/usr/lib/build/checks/* ; do
-	echo "... running ${CHECKSCRIPT##*/}"
-	$CHECKSCRIPT || cleanup_and_exit 1
+    declare -A check_script_schedule
+    for n in $BUILD_ROOT/usr/lib/build/checks/* ; do
+        if [[ "$n" =~ .*/([^-]+)-[^/]* ]]; then
+            check_script_schedule[${BASH_REMATCH[1]}]+=" $n"
+        else
+            cleanup_and_exit 1 "post-build check script name $n lacks a recognizable ordering prefix"
+        fi
+    done
+    indices=( ${!check_script_schedule[@]} )
+    IFS=$'\n' indices=( $(echo -e "${indices[@]/%/\\n}" | sed -r -e 's/^ *//' -e '/^$/d' | sort) )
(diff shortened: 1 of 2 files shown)


lnussel:
why not run scripts with the same prefix number in parallel like the old sysv startpar did?

peti:
Hmm, good point. We could do that, too. It's more complicated to implement in `bash`, of course, but it's certainly more expressive that what this PR implements now.

adrianschroeter:
The implementation seems not to be fully backward compatible. We can do that only for new distros, I think.

simotek:
We did look at doing a backward compatible implementation but ended up at this one as it was cleaner, I wondered if it would be a issue, if it is we can certainly do something slightly different that is compatible.

peti:
I will submit an updated version of this PR soon that will be backward-compatible. Just need to do some more testing ....

peti:
I pushed a new version that implements the behavior suggested in https://github.com/openSUSE/obs-build/pull/359#issuecomment-305121798. Scripts that have the same run-level will run in parallel. With https://github.com/openSUSE/post-build-checks/pull/10 applied, this will run almost all post-build checks in parallel. Without that change applied, all post-build checks will run sequentially, just like before, so this patch is safe to apply to `build` without having to apply https://github.com/openSUSE/post-build-checks/pull/10 at the same time.

peti:
Ping?

adrianschroeter:
The additional dependency to gnu_parallel is some problem, because we can not release maintenance updates easily anymore.

Is there an easy way to avoid this for you?

adrianschroeter:
eg. make it a recommend, since this dependency would be anyway ignored inside of VM builds.

So it would anyway break builds if you don't have any fallback implementation and the prjconf in OBS projects got not updated.

peti:
> Is there an easy way to avoid this for you?

No, the use of `parallel` is pretty much at the heart of what this PR does.

adrianschroeter:
hm, we will need to rewrite a bit then to have at least a fallback (we can not break all builds in OBS if distro config get not adapted).

Will try to discuss it here and take care next week ...

mlschroe:
Just don't require it and use the old code if parallel is not available. Newer distros can have a "Suggests: gnu-parallel" in the project config.

simotek:
I think checking if parallel is available and only using it when it is should be reasonable, I was about the suggest a similar thing.

peti:
@simotek, will you make the necessary changes?

simotek:
@peti I'm at a conference all next week so it will likely be quicker if you do.

peti:
@simotek, I think "quickness" is not a major concern for this PR any more. :-) If we'll have a fix in the next couple of weeks, then that's just fine IMHO. You have push access to the https://github.com/peti/obs-build repo, so if you could please implement your suggestion from https://github.com/openSUSE/obs-build/pull/359#issuecomment-359144206 and update the PR in the `master` branch there accordingly, then that would be great!

PR #363, opened 2017-06-13
Do not test debian changes file if not present

nickbroon:
Only run lintian if the changes file is present. This might not be the
case if for example a build.script is used that does not generate one
but still successfully completed, which should not cause lintian to
subsequently fail the build. lintian should only fail the build when it 
detects error in the changes.


nickbroon:
@gollub @jblunck

jblunck:
What do you mean by 'a build.script was used'. AFAIK the changes files should always get created.

nickbroon:
@jblunck 
The user can provide a custom "build.script" that could in theory do anything, which may or may one result in changes file. See https://github.com/openSUSE/obs-build/blob/master/build-recipe-dsc#L99 for where is this used.

jblunck:
That build.script is suppose to run dpkg-buildpackage, which itself is suppose to create and check the Debian changes file.

I don't know what exactly you want to achieve. Please explain your use-case a bigger detail.

nickbroon:
I would agree that default build cmd should perhaps be to call `dpkg-buildpackage -us -uc --check-command=lintian` instead of having the step afterwards calling lintian if the build cmd succeeds.

I've occasionally created a build.script for a package during development that does not actually create the debian package, but for example instead runs a static analyser/compiler, or just runs some tests.

It would thus be preferable if the recipe did not assume there will always be a .changes file generated by the build cmd.

As an alternative perhaps the recipe could check if lintian is present and if so add `--check-command=lintian` to the default build step?

nickbroon:
Bump. Any chance of approving/merging this?

nickbroon:
Bump. Any chance of approving/merging this?

nickbroon:
@adrianschroeter Can this be merged?

PR #366, opened 2017-06-19
vc - do not break hardlinks

sleep-walker:
some packages like kernel use hardlinks for changelogs

vc breaks it - this patch address it

### DIFF
--- a/vc
+++ b/vc
@@ -155,6 +155,6 @@ if [ -z "$message" ]; then
 fi
 mode=`stat -c "%a" "$changelog"`
 user=`stat -c "%u:%g" "$changelog"`
-mv "$tmpfile" "$changelog"
+cat "$tmpfile" > "$changelog"
 chmod $mode "$changelog"
 chown $user "$changelog"


adrianschroeter:
hm, don't like it much since it is no atomic operation anymore.

Also, where do you use vc? OBS interface does not support hardlinks either, so where would this be useful?

sleep-walker:
1. It wasn't much atomic operation before unless you have /tmp in same filesystem as your package. If you know how to atomically update hardlink, please tell me and I'll update the patch. Otherwise this fixes bug which is not corner case compared to requirement of atomic operation.
2. vc is called from `osc vc` which is used quite a lot
3. As I already said, kernel-* packages use that for changelogs in all SLE products.

sleep-walker:
I have to investigate, which part of the tools we use is responsible for doing hardlinks.

sleep-walker:
> Also, where do you use vc? OBS interface does not support hardlinks either, so where would this be useful?

It seems that hardlinking is done on our side for the changelogs - if the changelogs are identical on package creation, we make them hardlinks. After that editing one causes editing all.

PR #367, opened 2017-06-29
Add kiwi feature check method

schaefi:
Provide a generic method to detect the kiwi version. Based
on the version information provide a method to check for the
zsync feature. Further feature check methods can make use of
the same kiwi version detection method

### DIFF
--- a/build-recipe-kiwi
+++ b/build-recipe-kiwi
@@ -507,6 +507,34 @@ is_legacy_kiwi() {
     return 0
 }
 
+get_kiwi_version() {
+    # lookup version number of installed kiwi version and
+    # turn it into a reliable number
+    local version
+    local major
+    local minor
+    local patch
+    if is_legacy_kiwi; then
+        version=$(
+            chroot "$BUILD_ROOT" /usr/sbin/kiwi --version |\
+            sed -n 's,.*vnr: \(.*\),\1,p'
+        )
+    else
+        version=$(
+            chroot "$BUILD_ROOT" kiwi-ng --version |\
+            sed -n 's,.*version \(.*\),\1,p'
+        )
+    fi
+    major=$(printf %d $(echo $version | cut -f1 -d .))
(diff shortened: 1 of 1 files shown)


build-recipe-kiwi
```diff
@@ -507,6 +507,34 @@ is_legacy_kiwi() {
...
 }
 
+get_kiwi_version() {
+    # lookup version number of installed kiwi version and
+    # turn it into a reliable number
+    local version
+    local major
+    local minor
+    local patch
+    if is_legacy_kiwi; then
```
Conan-Kudo: How does it know it's a legacy kiwi? I don't see anything in the function that tells it that it is...

If anything, the simpler thing to do would be the reverse the logic and check for `kiwi-ng` if it exists, and if it does, you know it's not legacy kiwi, then check the version. If the version is high enough, it's good for zsync support check.

schaefi: There is a is_legacy_kiwi method in build/build-recipe-kiwi, existing for some time and already in use. Thus I think it's not advisable to change the working part ;)

adrianschroeter:
kiwi_has_zsync_support() is still not used as far as I see. Is there code comming using it? Or why do we need this code here?

schaefi:
It's almost 5month in the past but iirc and per issue https://github.com/SUSE/kiwi/issues/364 it was the buildservice who suggested that the kiwi bundler should be able to create a zsync bundle.
Unfortunately it was me who opened the issue as the result of a communication almost half a year ago. Thus the history of the originator is lost.

This pull request was a, let's say service from the kiwi team to provide the buildservice a method to check if the used kiwi builder has the zsync capability or not, because I thought you will need that when you add zsync image bundle support in obs.

Thus I think you can best answer your own question. If zsync image bundle is something you want, this pull request could be of help, else it could also be closed.

In general I think it is useful if the buildservice is able to check for builder capabilities according to the version as we had this problem with other features in the past too

mlschroe:
Closing as suggested.

PR #368, opened 2017-07-06
Debian: fix conflict with Debian's osc and obs-build

bluca:
Debian ships build in a package named obs-build. Given the package names are different, but ship the same files, there is a conflict on the file system. This was sent via https://build.opensuse.org/request/show/450943 but it got lost when the files moved into git.

Also add symlinks so that build can work with Debian's version of osc out of the box.

### DIFF
--- a/dist/debian.control
+++ b/dist/debian.control
@@ -5,10 +5,13 @@ Maintainer: Adrian Schroeter <adrian@suse.de>
 Build-Depends: debhelper (>= 4)
 Standards-Version: 3.7.2
 
-Package: build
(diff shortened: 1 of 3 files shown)


adrianschroeter:
hm, shouldn't we rename the package name in our builds then?

bluca:
I can rename and take away the conflict/replace/provides if you are ok with that, give me a few mins

bluca:
the symlinks will still be needed for full compatibility

bluca:
@adrianschroeter done.

PR #372, opened 2017-08-01
Point fissile to the cache dir, save generated images and exit on fai…

jimmykarily:
…lures


jimmykarily:
Thanks for merging, let me know when you deploy this

PR #374, opened 2017-08-04
Instruct fissile to use --network=host when building packages

jimmykarily:
otherwise we get errors because "localhost" can't be resolved to "127.0.0.1"


jimmykarily:
Example log here: https://build.opensuse.org/package/live_build_log/Cloud:Platform:scf:fissile-releases/uaa/openSUSE_Leap_42.2/x86_64

PR #375, opened 2017-08-11
Define SOURCEURL and PATCHURL macros

marcus-h:


mlschroe:
What would you do with that INT_MAX  case? I don't see how this can be reliable used in a spec file.

marcus-h:
On 2017-08-31 12:09:55 +0000, Michael Schroeder wrote:
> What would you do with that INT_MAX  case? I don't see how this can be reliable used in a spec file.
> 
Me neither (at least it is not portable)... but that's what rpm does
(see addSource(...) in build/parsePreamble.c). I haven't checked why
this "strange" handling of an unnumbered patch is needed...

mlschroe:
Merged without the INT_MAX...

PR #381, opened 2017-08-29
determine command to start a shell in kvm / docker

lethliel:
determine command to start a shell in kvm / docker and write it into the buildroot. 

The command is written to:

- .get_shell_docker for docker
- .get_shell_kvm for qemu / kvm 
- more to follow

These files will later be used to provide a shell in the desired build container.


lethliel:
not needed anymore

PR #386, opened 2017-09-28
Move debian directory to different namespace

zumbi:
Official Debian packaging creates debian directory, we would like
  to keep upstream sources and Debian packaging in same source
  repository, however having a debian directory in upstream source
  difficults packaging work.

Signed-off-by: Héctor Orón Martínez <zumbi@debian.org>


Conan-Kudo:
@zumbi Why are you merging your packaging with the source tree? Just keep the `debian/` directory in its own repository and reference upstream git when releasing builds to Debian. I do something similar for Fedora packaging.

adrianschroeter:
did it in a different way and adapted _service files where needed.

PR #387, opened 2017-10-04
vc: prevent leading newline when using the just_edit mode

lethliel:
when using vc in just_edit mode a newline is added at the beginning of the file.


lethliel:
Looks good to mls

PR #388, opened 2017-10-05
Only block .chroot archive config

nickbroon:
.binary config seems reasonable to allow user to add, as it effect the
final live system.
.chroot is for the livebuild chroot, where there is no network access,
during build on OBS so blocking them seems reasonable.


nickbroon:
@gollub @jblunck

PR #391, opened 2017-10-17
Facilitate debugging of the build system scripts call chain

jimklimov:
build build-recipe init_buildsystem : report details of this script's startup into the log

### DIFF
--- a/build
+++ b/build
@@ -25,6 +25,9 @@
 #
 ################################################################
 
+echo "`date -u` : starting $0 $*" >&2
+ps -ef | grep -vw ps | grep -vw grep | grep -w $PPID
+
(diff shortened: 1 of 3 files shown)


adrianschroeter:
sorry, this looks more like temporary debugging hacks then something what should be shown on each build ...

jimklimov:
Well, sort of - it is indeed helpful to debug the (on-premise) OBS setup especially when creating new jobs. It does not add much overhead in the logs so we kept it in production. Possibly because the on-premise production setup still undergoes experimental improvement, so quickly seeing in the logs that a build went as expected - or where it went wrong early on - remains a useful feature.

PR #392, opened 2017-10-17
Speed up debian packaging with parallel builds and deferred pkg trigger processing

jimklimov:
Parallel: older compat-level packages do not default to parallel builds and should be nudged, both in the `dpkg*` command call and in the rules-file. So far this is a valid use-case for some on-premise deployments and projects that should support older debian-like distributions.

Deferred trigger - mainly man-db is a single big offender in our setups instantiating of the build root. Doing it once, rather than for every other package during its installation (takes 3-6 sec each on our ARM workers) gives noticeable benefit since this process is forked off a lot less, at least (does take a minute for the single chunk of more to process in one bite, though).

### DIFF
--- a/build-pkg-deb
+++ b/build-pkg-deb
@@ -58,7 +58,7 @@ pkg_initdb_deb() {
 	rm -f $BUILD_ROOT/.init_b_cache/dpkg.deb
 	cp $BUILD_ROOT/.init_b_cache/rpms/dpkg.deb $BUILD_ROOT/.init_b_cache/dpkg.deb || cleanup_and_exit 1
     fi
-    deb_chroot $BUILD_ROOT dpkg --install --force-unsafe-io --force-depends .init_b_cache/dpkg.deb >/dev/null 2>&1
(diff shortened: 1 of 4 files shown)


nickbroon:
Some of the Lintian configurable toggle changes overlap with #363 which only runs lintian of there is a changes files, as jobs might override the build command and not generate one.

bluca:
@jimklimov as requested on IRC:

> (10:57:33) bluca: jimklimov1: to skip lintian in a debian build, you can just avoid to include it in the build
(10:57:50) bluca: the dsc recipe won't run it if it's not there, and won't fail either
(10:58:05) bluca: much simpler than having to change the code and setting configs, etc etc
(10:59:06) bluca: (just seen your obs-build PR)
(11:00:27) bluca: and if your global prjconf includes it, in a project where you don't want it to get installed you can just add "Support: !lintian" to its prjconf
(11:01:02) bluca: (I do that all the time when I have to backport from unstable, as often there are false positives corrected in later versions)
(11:02:41) jimklimov: @bluca : thanks, good hint :) can you plz put it in the ticket comments, so I don't forget it in IRC (on commute now)? and perhaps so nobody merges the commit if it is not needed after all...

PR #396, opened 2017-10-19
Improve usage of ccache to speed up builds

jimklimov:
Port of our (older) OBS setup, which benefits greatly from use of the same persistent `ccache` directory on all workers, ARM and amd64 alike (and Jenkins workers running builds in similar OS images too, added for good measure).

These changes allow to optionally enforce usage of `ccache` binary provided by the worker host (such approach should suffice for most Linux distros) even if the build-root does not pull ccache as a package, and to optionally bind-mount the cache directory from the worker's host so that a shared cache is easy to do.

### DIFF
--- a/build
+++ b/build
@@ -103,6 +103,7 @@ SIGNDUMMY=
 DO_STATISTICS=
 RUN_SHELL=
 CCACHE=
+#DO_CCACHE_STATISTICS=
 DLNOSIGNATURE=
 BUILD_FLAVOR=
 OBS_PACKAGE=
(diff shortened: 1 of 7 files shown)


AdamMajer:
ping? Anyone looking at this? Having ccache available in a project (or per project) would be a great benefit considering we have plenty of "rebuilds" where nothing changes.

adrianschroeter:
Sumit reworked the ccache a bit. It is merged meanwhile Could you check if this sufficient for you?

PR #398, opened 2017-10-19
Deb.pm: support Build-Depends-Arch

suihkulokki:
Build-Depends-Arch and Build-Conflicts-Arch where added in dpkg 1.16.4.
They are now documented in man deb-src-control(5) and debian policy.
Add support to obs-build as packages in unstable are starting to use
them.


mlschroe:
Thanks a lot!

PR #400, opened 2017-11-13
Keep preinstalled packages

### LINKED ISSUE #377: switching rpm versions breaks build
When building with kvm without --clean
switching between different rpm binaries (with unimportant patch differences) breaks the build with a message like
`chroot: failed to run command 'rpm': No such file or directory`

Here is a reproducer:
```bash
#!/bin/sh
# random package, chosen because build is fast
p=openSUSE:Factory/skelcd-openSUSE
osc co $p
cd $p
# two random projects that have different rpm binaries
o='osc build --vm-type=kvm'
$o --alternative-project=home:bmwiedemann:reproducible --clean
$o --alternative-project=home:bmwiedemann:reproducible:test
```

bmwiedemann:
when building in KVM, we need the preinstalled packages (e.g. rpm)
to install packages, because once we uninstall them, we cannot get them back.

Closes #377


jimklimov:
LGTM, makes sense

PR #410, opened 2017-11-27
[build] fixed typo in build-vm-docker

mmohring:
syntax error in line 52 of build-vm-docker


mmohring:
@eduardoj that was fast 👍

PR #411, opened 2017-11-28
Use docker --mount instead of --volume

marcus-h:
Completely _untested_ (based on the information in
https://docs.docker.com/engine/admin/volumes/volumes/#differences-between--v-and---mount-behavior)

This is needed to support colons in the volume name.

Fixes: https://github.com/openSUSE/osc/issues/358 ("--vm-type=docker
not compatible with build_root with common prj names containing ':'")

### DIFF
--- a/build-vm-docker
+++ b/build-vm-docker
@@ -32,7 +32,7 @@ vm_startup_docker() {
     docker rm "$name" >/dev/null 2>&1 || true
     docker run \
         --rm --name "$name" --cap-add=sys_admin --cap-add=MKNOD --net=none \
-        -v "$BUILD_ROOT:/mnt" busybox /bin/chroot /mnt "$vm_init_script"
+        --mount "type=volume,source=$BUILD_ROOT,destination=/mnt" busybox /bin/chroot /mnt "$vm_init_script"
     BUILDSTATUS="$?"
     test "$BUILDSTATUS" != 255 || BUILDSTATUS=3
     cleanup_and_exit "$BUILDSTATUS"


nickbroon:
:+1:

nickbroon:
It appears that docker also doesn't like `:` in mount names either. :-(
I think that obs-build/osc really will need to replace : with something like _ as discussed in openSUSE/osc#358

```
[    0s] init_buildsystem --configdir /usr/lib/build/configs --cachedir /var/cache/build --prepare --rpmlist /var/tmp/osc-buildpackage/rpmlist.JCmlBe /var/tmp/osc-buildpackage/tmp.YWXfdxNcpr/package_1.2.3.dsc ...
[    0s] copying packages...
[    1s] reordering...cycle: libc6 -> libgcc1
[    1s]   breaking dependency libgcc1 -> libc6
[    1s] cycle: dh-python -> python3
[    1s]   breaking dependency python3 -> dh-python
[    1s] cycle: debhelper -> dh-autoreconf
[    1s]   breaking dependency debhelper -> dh-autoreconf
[    1s] cycle: debhelper -> dh-strip-nondeterminism
[    1s]   breaking dependency debhelper -> dh-strip-nondeterminism
[    1s] cycle: dmsetup -> libdevmapper1.02.1
[    1s]   breaking dependency libdevmapper1.02.1 -> dmsetup
[    1s] done
[    2s] booting docker...
[    2s] docker: Error response from daemon: create /var/tmp/osc-buildpackage/build-root/home:user:package-Debian_9_Update_standard-x86_64/.mount: "/var/tmp/osc-buildpackage/build-root/home:user:package-Debian_9_Update_standard-x86_64/.mount" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
```

mlschroe:
But shouldn't that be type=bind instead type=volume in the mount option?

nickbroon:
Yes. With the following change I got a build to work.

``` diff
@@ -32,7 +32,7 @@
     docker rm "$name" >/dev/null 2>&1 || true
     docker run \
         --rm --name "$name" --cap-add=sys_admin --cap-add=MKNOD --net=none \
-        -v "$BUILD_ROOT:/mnt" busybox /bin/chroot /mnt "$vm_init_script"
+        --mount "type=bind,source=$BUILD_ROOT,destination=/mnt" busybox /bin/chroot /mnt "$vm_init_script"
     BUILDSTATUS="$?"
     test "$BUILDSTATUS" != 255 || BUILDSTATUS=3
     cleanup_and_exit "$BUILDSTATUS"
```

mlschroe:
Great, thanks. Merged and changed to type=bind.

PR #418, opened 2018-01-04
Fix rpm detection in build-recipe-appimage

marcus-h:
Check for the presence of rpm in the build root instead of the
host system.

Fixes: https://github.com/openSUSE/osc/issues/365 ("cannot build
sample AppImage with osc on ubuntu 16.04")


marcus-h:
Superseded by https://github.com/openSUSE/obs-build/pull/419

PR #419, opened 2018-01-04
Fix rpm detection and createrepo code in build-recipe-appimage

marcus-h:
Check for the presence of rpm in the build root instead of the
host system. Execute createrepo in the build root and not in
the host system.

Fixes: https://github.com/openSUSE/osc/issues/365 ("cannot build
sample AppImage with osc on ubuntu 16.04")


adrianschroeter:
ups, thanks a lot!

PR #421, opened 2018-01-15
Fix sysdeps parsing in case of kiwi image

aaannz:
- Build::get_sysbuild returns and array where at the first position
is the result of the whole operation. This was not handled properly
for kiwi outputs


aaannz:
Anything I can do to make this merged? Without this change, when build script is called from command line I get this result:
```
[    2s] Use of uninitialized value within %packs in hash element at /usr/lib/build/expanddeps line 353.
[    2s] Use of uninitialized value in concatenation (.) or string at /usr/lib/build/expanddeps line 353.
[    2s] Use of uninitialized value within %packs in exists at /usr/lib/build/expanddeps line 354.
[    2s] unsupported url for '1':
```

PR #422, opened 2018-01-17
[bugfix] avoid warnings for %%foo

M0ses:


M0ses:
Looks like the problem is somewhere else

Issue #425, opened 2018-01-23
Provide a more useful definition for _smp_mflags override


StefanBruens:
build-recipe-spec defines _smp_mflags as:

`%_smp_mflags -j$BUILD_JOBS`

while the default rpm definition is:
```
%_smp_mflags %([ -z "$RPM_BUILD_NCPUS" ] \\\
        && RPM_BUILD_NCPUS="`/usr/bin/getconf _NPROCESSORS_ONLN`"; \\\
        ncpus_max=%{?_smp_ncpus_max}; \\\
        if [ -n "$ncpus_max" ] && [ "$ncpus_max" -gt 0 ] && [ "$RPM_BUILD_NCPUS" -gt "$ncpus_max" ]; then RPM_BUILD_NCPUS="$ncpus_max"; fi; \\\
        if [ "$RPM_BUILD_NCPUS" -gt 1 ]; then echo "-j$RPM_BUILD_NCPUS"; fi)
```

The important difference is it allows limiting the number of jobs using %_smp_ncpus_max. This is especially useful when the job count is too high for the available memory.

See e.g. http://opensuse.14.x6.nabble.com/Firefox-builds-on-OBS-fail-td5057557.html

mlschroe:
We now changed to code so that it makes use of RPM_BUILD_NCPUS if available.

PR #426, opened 2018-01-23
Allow limitting number of jobs using _smp_ncpus_max macro

StefanBruens:
The default definition for _smp_mflags allows limiting the number of
jobs by specifying _smp_ncpus_max, which is required in memory constrained
environments.

Fixes: #425 ("Provide a more useful definition for _smp_mflags override")

### DIFF
--- a/build-recipe-spec
+++ b/build-recipe-spec
@@ -80,10 +80,21 @@ recipe_prepare_spec() {
 ' >> $BUILD_ROOT/root/$rawcfgmacros
     fi
 
-    if test -n "$BUILD_JOBS" ; then
+    if test "$icecream" -gt 0 ; then
+        # hardcode number of jobs, icecream workers can not be probed from the build environment
 	cat >> $BUILD_ROOT/root/$rawcfgmacros <<-EOF
 		%jobs $BUILD_JOBS
 		%_smp_mflags -j$BUILD_JOBS
+		EOF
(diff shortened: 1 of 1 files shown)


adrianschroeter:
I am not really a fan of having such a macro hardcoded in build script .... can't you submit that to rpm package instead? Who says that this bash syntax will stay to work on long time frame?

adrianschroeter:
This just does not belong to the build script, so I close it here.

Feel free to submit it to the rpm package of the distro. Or create your own package providing it via/usr/lib/rpm/macros.d/macros.XXXX . You should however avoid then that always a bash gets forked also when max macro is not used at all.

StefanBruens:
This macro *is* essentially the macro from rpm itself (e.g. /usr/lib/rpm/platform/x86_64-linux/macros), but the build script *hardcodes* it to some different value.

adrianschroeter:
k, still, we don't want to have a concrete implementation of it in the build script since this needs to stay compatible forever. While macros inside of the packages and architectures can be modified and developed without breaking old builds. So replacing it via some package is the more safe way here ...

StefanBruens:
Can you clarify what you mean by "replacing it via some package?"
I want to rely on this, so a package can influence the maximum number of jobs. Currently a package can do this when build using plain rpmbuild, but this breaks as soon as the build script is used.

PR #429, opened 2018-02-10
Don't rely on AppImage's auto-detection for architecture

Martchus:
Otherwise build errors occur when AppImage's auto-detection for the architecture fails, eg.:
```
[  367s] Use appimagetool
[  367s] /usr/src/packages/OTHER /usr/src/packages/SOURCES
[  367s] appimagetool, continuous build (commit 88400e4975cd3fb1270e5a6352a7180195dfae83), build (null) built on 
[  367s] Unable to guess the architecture of the AppDir source directory "/usr/src/packages/Tag Editor-0-Build5.1-x86_64"
[  367s] A valid architecture with the ARCH environmental variable should be provided
[  367s] e.g. ARCH=x86_64 appimagetool ...
[  367s] 
[  367s] cloud136 failed "build appimage.yml" at Sat Feb 10 19:42:41 UTC 2018.
```

I tested locally that passing the architecture manually like this PR proposes works, at least for `x86_64`.

Note that the error message is emitted here: https://github.com/AppImage/AppImageKit/blob/appimagetool/master/src/appimagetool.c#L607

As a workaround I could just place an `*.so` file in the directory the auto-detection checks, but I think it is cleaner not to rely on the auto-detection and just pass the architecture since it is determined anyways.

### DIFF
--- a/build-recipe-appimage
+++ b/build-recipe-appimage
@@ -91,7 +91,7 @@ recipe_build_appimage() {
         release_option="--release $RELEASE"
     fi
 
-    chroot $BUILD_ROOT su -c "cd $TOPDIR/SOURCES && /usr/lib/appimagetool/pkg2appimage $release_option" - root \
+    chroot $BUILD_ROOT su -c "cd $TOPDIR/SOURCES && ARCH=$ARCH /usr/lib/appimagetool/pkg2appimage $release_option" - root \
 	|| cleanup_and_exit 1
 
     # extract build result basenames


Martchus:
Any feedback?

adrianschroeter:
I don't mind to take it, but I would like to hear an opinion from AppImage people if the problem should be really solved here.

@TheAssassin @probono any opinion?

TheAssassin:
At the moment, appimagetool tries to detect which architecture an app has before it gets packaged as an AppImage. This auto detection relies on `.so` and executable files in the AppDir (the "source" directory for the AppImage build) and isn't perfect therefore. For instance, if you have a mix of 32-bit and 64-bit executables (which is perfectly valid, e.g., totally static binaries such as busybox may be shipped as 32-bit binaries to run universally on all x86 architectures), appimagetool can't tell the architecture reliably.

This string is IIRC only used for generating the filename, and, thinking about this code, it doesn't make any real sense to do it that way (especially since the runtime in a 32-bit AppImage that would be named `i686` therefore but built with `appimagetool-x86_64.AppImage` would still carry the 64-bit runtime).
Therefore, I'd suggest you `export ARCH=(x86_64|i686|...)`, as @Martchus suggested.

@Martchus I'd appreciate if you could open an issue about this section in our code to trigger a discussion on finding a better method. I think there might even be an issue about that in our tracker already...

TL;DR: Please merge this change.

PR #432, opened 2018-02-28
Fixed use of provides tags

schaefi:
A provides tag of the format "foo:bar" causes a problem in rpm if the macro:

* %_wrong_version_format_terminate_build

is not set to 0. For suse builds the above macro is set to 0 and thus do not cause problems. But for other distributions the build fails if the kiwi packages creates provides tags in this format.

Therefore this patch is one out of series which changes the use of the provides tags to the format "foo(bar)". There will be reference patches to the spec files of the

* python-kiwi
* kiwi-boot-descriptions

packages to complete the required changes.

If you agree on the change I will put in the reference pull requests once I have created them so that all changes can be merged in one shot

Thanks

### DIFF
--- a/Build/Kiwi.pm
+++ b/Build/Kiwi.pm
@@ -97,7 +97,7 @@ sub kiwiparse {
       if (defined $type->{'image'}) {
         # for kiwi 4.1 and 5.x
         push @types, $type->{'image'};
-        push @packages, "kiwi-image:$type->{'image'}" if $schemaversion >= $schemaversion56;
+        push @packages, "kiwi-image($type->{'image'})" if $schemaversion >= $schemaversion56;
       } else {
         # for kiwi 3.8 and before
         push @types, $type->{'_content'};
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
This change helps me out considerably in Fedora, and I'll probably pull this into my Fedora packaging of `obs-build` (corresponding with my Fedora packaging for `kiwi`, which will use this Provides/Requires form).

adrianschroeter:
This would break all builds using the old kiwi packages, no?

Conan-Kudo:
@adrianschroeter It probably would. It might make sense to adjust this so that it checks for new style Provides first, and then falls back to old-style ones.

mlschroe:
Wait wait wait, all this doesn't make much sense to me. The _wrong_version_format_terminate_build rpm flag makes rpm fail if there is a ':' in the *version* part of a provides. But there are no versions involved here.

mlschroe:
Anyway you'd never change the parser. The original intend was to use a Substitute in the project config to map the dependency into something known in the distribution. So closing this here.

PR #437, opened 2018-04-12
Kiwi build doesn't work because of expanddeps crash

illuusio:
This one fixes Kiwi build with error:
[    0s] expanding package dependencies...
[   18s] Use of uninitialized value within %packs in hash element at /usr/lib/build/expanddeps line 354.
[   18s] Use of uninitialized value in concatenation (.) or string at /usr/lib/build/expanddeps line 354.
[   18s] Use of uninitialized value within %packs in exists at /usr/lib/build/expanddeps line 355.
[   18s] unsupported url for '1':

### DIFF
--- a/expanddeps
+++ b/expanddeps
@@ -420,7 +420,7 @@ Build::readdeps($cf, undef, \%repo);
 
 #######################################################################
 
-my @sysdeps = Build::get_sysbuild($cf, $buildtype, $extrasysdeps);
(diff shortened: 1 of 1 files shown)


aaannz:
You can't fix it like this, since if not kiwi build, later it shifts `@sysdeps` expecting the 1 or 0 result. See https://github.com/openSUSE/obs-build/pull/421 (which is without response 3 months now. :/ )

illuusio:
Which line because I didn't spot it

Issue #439, opened 2018-04-27
changelog2spec 'messes' with timezones


DimStar77:
Take this very simple .changes file:

```
-------------------------------------------------------------------
Fri Apr 28 03:00:00 CEST 2018 - dimstar@opensuse.org

- Entry2

-------------------------------------------------------------------
Fri Apr 27 23:59:59 UTC 2018 - dimstar@opensuse.org

- Entry1
```

When parsing this through changelog2spec, the result is:

```
* Sat Apr 28 2018 dimstar@opensuse.org
- Entry2
* Sat Apr 28 2018 dimstar@opensuse.org
- Entry1
```

The 2nd entry now magically mecame the 28. of Apirl - but since there is no timezone information, this is actually wrong. If already normalzing times, then I'd strongly suggest NOT to do it on the 'timezone in the most recent entry' - but on UTC (to get a consistent conversion; if multiple users update the changes file using different timezones, the dates keep on swapping around)

PR #440, opened 2018-04-27
Set target timezone to UTC (offset 0)

DimStar77:
Instead of normalizing all dates/times in the .changes file to
whatever timezone the first entry has, let's normalize to UTC, especially
as this information is not found in the rpm changelog in the end.

Fixes issue #439.


Conan-Kudo:
I'd actually like to see the timezone information encoded in the generated `%changelog` section from changes files. Since RPM 4.14, this is now supported with the same timestamp form that is used in changes files today.

DimStar77:
ping?

mlschroe:
What Neal said... we really should move to the new format if we detect a new enough rpm version.

PR #441, opened 2018-05-03
Fix rich dependency calculation when dependency name has parentheses

OlegGirko:
This change fixes `getnevr_rich()` function to work correctly with dependencies containing parentheses.

For example, this function was mistakingly returning `rubygem(xpath` when parsing the following rich dependency:
`(rubygem(xpath) >= 2.0 with rubygem(xpath) < 3)`


mlschroe:
Thanks! I've also fixed another bug: it could not handle two '(' in a row.

PR #444, opened 2018-05-31
Do not pass the deprecated "--force" option to pacman -U

marcus-h:
Instead, pass --overwrite '*' to pacman -U, which has the same effect
as the deprecated --force option. Note: this changes assumes that
the pacman -U command supports the --overwrite option (which is ok,
because Arch is a rolling distro).

If you think we should care about backward compatibility with old pacman
versions, I'll happily update the PR:)


adrianschroeter:
seems we don't need to care about stability on arch, since there is no older stable release...

PR #445, opened 2018-06-01
[archlinux] replace pkg-config with pkgconf

mmetak:
https://lists.archlinux.org/pipermail/arch-dev-public/2018-May/029252.html


adrianschroeter:
Thanks, I have also adapted the config in Arch:Core project on build.opensuse.org
(the config in this package is only for builds with plain build script)

PR #448, opened 2018-06-10
computeblocklists: add fallback code if the "Q" type is not supported

marcus-h:
Since the "Q" type is not available on all platforms or perl builds,
fallback code is needed in these cases. For this, instead of using
the unsigned quad type ("Q"), use two unsigned long (32-bit) types
("L").
To pack an unsigned quad, it is "halved" into two unsigned longs
(32-bit). Afterwards, the two unsigned longs are packed via "LL"
(taking the endianness into account (either little-endian or
big-endian (no "weird" byte orders))).
To unpack an unsigned quad, two unsigned longs (32-bit) are unpacked
and then represented by a Math::BigInt instance (again, taking the
endianness into account). Using a Math::BigInt potentially slows down
arithmetic operations.

This is a follow-up fix for commit 8c90847 ("computeblocklists: add
support for FIEMAP ioctl").

Note: it is not guaranteed that the "extractbuild" script can deal
with the results of the "computeblocklists" script. For instance,
assume:
- The perl that is used to execute the "extractbuild" script does not
  support 64-bit integers
- We have to read a block that cannot be represented by a 32-bit
  integer. Also, it cannot be represented as a perl floating point
  number (see "perldoc perlnumber").

Hence, most of the arithmetic that is performed with this block in
"extractbuild" results in "garbage". Using Math::BigInt might help
here, but then we would need to check if, for instance, a
seek(F, $some_big_int, 0) is translated to a _llseek/whatever...

@jeffmahoney or someone else: if there is a more
straightforward/elegant solution please tell me:)

### DIFF
--- a/computeblocklists
+++ b/computeblocklists
@@ -32,6 +32,7 @@
 ################################################################
 
 use strict;
+use Math::BigInt;
 
 # powerpc, mips, sparc, and alpha have 3 direction bits
 # and represent _IOC_NONE as 1 instead of 0.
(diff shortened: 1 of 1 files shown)


mlschroe:
The 56bit mantissa of a perl float should be good enough for our use...

mlschroe:
- I don't think we need to use bigint
- there's a much more elegant way to do it codewise. I'll commit something after doing some tests on some weird architectures we have (armv7, s390, ppc)

mlschroe:
Hopefully fixed with commit 4be1109315cf32d849af2912fb9702428743315c.

PR #450, opened 2018-06-28
Enable Core dump generation in VM

ggardet:


lnussel:
not good. leads to core dump files in random directories:
https://bugzilla.suse.com/show_bug.cgi?id=1103370

What is the justification for this feature? Why are core dumps in random directories needed?

PR #452, opened 2018-07-25
Disable kpti on guest kernel for performance and correctness

dirkmueller:
kpti costs performance for VM guests but doesn't bring security
benefits (the guest kernel shall be considered untrusted anyway
because the user can define a custom kernel for his project).
As a side effect it avoids a regression on some ARM SoCs (bsc#1097449)


plinnell:
LGTM

adrianschroeter:
moved it to the generic, not KVM specific definition.

PR #453, opened 2018-08-03
build-recipe: Adjust check for email only authors to be specific for SLE

Conan-Kudo:
The way the condition was structured, setting `%fullname_in_changelog` to `1` would trigger the wrong behavior on non-SUSE systems.

This conditional evaluates to disable full identities only if it is a SUSE system that isn't openSUSE only if `%fullname_in_changelog` is not set.

In all other cases, it preserves full identities if they exist.

### DIFF
--- a/build-recipe
+++ b/build-recipe
@@ -163,10 +163,10 @@ recipe_create_changelog() {
     esac
     # add --emailonly option for sles builds
     if test "$CFFORMAT" = rpm ; then
-	case `queryconfig --dist "$BUILD_DIST" --configdir "$CONFIG_DIR" --archpath "$BUILD_ARCH" eval '%{?is_opensuse}/%{?!fullname_in_changelog:%{?suse_version}}'` in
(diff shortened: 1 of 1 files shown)


mlschroe:
Hmm, I don't see why the code as is doesn't work. For non-suse systems with fullname_in_changelog set to 1 you'll get '/' after maco evaluation, which does not match "*/*[1-9]*". So --emailonly will not be used.

(But I admit that I should simply have used `%{?is_opensuse}%{?fullname_in_changelog}/%{?suse_version}` instead)

The reason why there's a case statement is that we'll also want full changelogs in future sles versions.

mlschroe:
No, if you define `%fullname_in_changelog 1` the "right side" (right of the '/') of the expansion will always be empty, so you'll never get --emailonly.

Anyway, I'll simplify the code to the expansion I had in my last comment.

Conan-Kudo:
@mlschroe Ah, okay. Then sorry for the noise.

PR #456, opened 2018-08-29
debootstrap: ensure /etc/hosts exists

alee-ccu:
debootstrap won't necesserily create /etc/hosts, so copy it from the
main buildroot if it's not there.

Signed-off-by: Héctor Orón Martínez <zumbi@debian.org>

### DIFF
--- a/build-recipe-debootstrap
+++ b/build-recipe-debootstrap
@@ -78,6 +78,10 @@ recipe_build_debootstrap() {
 
     chroot $BUILD_ROOT/$myroot apt-cache gencaches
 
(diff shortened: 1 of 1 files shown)


alee-ccu:
Hi. I am `obs-build` package maintainer in Debian:

  https://sources.debian.org/src/obs-build/20180302-3/debian/patches/0002-debootstrap-ensure-etc-hosts-exists.patch/

alee-ccu:
There was some packages failed to build from sources due to it's tests failed to resolve `localhost`. . eg: glibc

adrianschroeter:
ah, forget my comment, it is only a copy inside the build env ...

PR #459, opened 2018-08-29
build-recipe-dsc: avoid source package rebuilds

alee-ccu:
If we use build suffixes, rebuilding the source package will
result in the build suffix being included into the source package,
which we'd rather avoid.

Attach patch to avoid source package rebuild, which causes
issues as the source package appends undesired suffix.

Author: Andrew Shadura <andrew.shadura@collabora.co.uk>

### DIFF
--- a/build-recipe-dsc
+++ b/build-recipe-dsc
@@ -95,7 +95,7 @@ dsc_build() {
     # this allows the build environment to be manipulated
     # and alternate build commands can be used
     DSC_BUILD_CMD="$(queryconfig --dist "$BUILD_DIST" --archpath "$BUILD_ARCH" --configdir "$CONFIG_DIR" substitute dsc:build_cmd)"
-    test -z "$DSC_BUILD_CMD" && DSC_BUILD_CMD="dpkg-buildpackage -us -uc"
+    test -z "$DSC_BUILD_CMD" && DSC_BUILD_CMD="dpkg-buildpackage -us -uc -b"
     if test -e $buildroot/$TOPDIR/SOURCES/build.script ; then
 	echo "Sourcing build.script to build - it should normally run 'dpkg-buildpackage -us -uc'"
 	DSC_BUILD_CMD="source $TOPDIR/SOURCES/build.script"


alee-ccu:
Hi. I am obs-build package maintainer in Debian:

https://sources.debian.org/src/obs-build/20180302-3/debian/patches/build-recipe-dsc-don-t-rebuild-the-source-package.patch/

adrianschroeter:
You speak about to avoid source package rebuilds, but this would disable any generation of source package at all, no?

alee-ccu:
The source package should stays as same as it's uploaded. We only need to generate binary packages with build suffixes.

Without this patch. When we build with suffixes, the source package gets suffixes appended each time when a package triggered rebuilds which is very annoying for building and maintaining a full distro.

adrianschroeter:
In which context are you using this? Because in OBS case there is no source package uploaded, just single source files. 
The source package gets build as a result of these source files. And you may even need multiple of them if you use Substitute functionality which is patching deps.
Or is this about direct usage of the build script itself?

bluca:
I think this would also break running Lintian on the source tarball:

https://github.com/openSUSE/obs-build/blob/188f3aa121a507972a311c5846b188dfffbda015/build-recipe-dsc#L111

Conan-Kudo:
This firmly breaks the reproducibility between binaries and sources. A produced binary package should have a corresponding produced source package that can be used independently of the build system to produce the package. By disabling source package builds, you've effectively stopped supporting that model.

smcv:
> In which context are you using this? Because in OBS case there is no source package uploaded, just single source files.

The way we're using OBS is that the Debian source package (the `.dsc` file, plus the tarballs that it lists) is what's committed to OBS, and the binaries are built from that. We aren't committing a directory containing `*.c` or anything like that. Is that the model you had in mind?

smcv:
> I think this would also break running Lintian on the source tarball

If the `.dsc` is what's gets committed, then it's that `.dsc` (if anything) that should be put through Lintian.

Running Lintian on the output `.changes` file still analyzes the contents of the binary packages (`.deb`).

smcv:
> This firmly breaks the reproducibility between binaries and sources. A produced binary package should have a corresponding produced source package that can be used independently of the build system to produce the package.

We do - but the source package is prepared elsewhere, and committed to OBS. When you `osc co` a package that is maintained like that, you get `.dsc`, `.orig.tar.*`, `.debian.tar.*` files, which can be unpacked with `dpkg-source` to get the editable source code.

(Strictly speaking the source that is compiled is not *identical* to the `.dsc` that was committed, because its `debian/changelog` is sometimes modified to include an extra top changelog entry for the `Release`; but that's the same as what's done in Debian itself when release managers do a "binNMU", in which an existing source package is rebuilt against updated dependencies, with no non-changelog changes. In Debian, binNMUs of a package `hello_1.2.3-4.dsc` would be versioned like 1.2.3-4+b1, 1.2.3-4+b2 and so on.)

Conan-Kudo:
@smcv I don't think you understand what I'm saying. I'm saying that there should be a source package that includes all of the OBS transformations on top of this. In no way am I condoning the idea that sources should be disconnected from the build process (which is what you are proposing).

sjoerdsimons:
@Conan-Kudo on the flip side you're missing what @smcv is saying somewhat; What this patch intends to do is to move it to the same model as Debian BinNMUs which is for pure recomplications without source changes in the Debian world (only a package version bump to reflect the rebuild). 

It doesn't seem unreasonable for OBS to support that model as well, though it should probably be configurable and *only* for packages with a source package in obs rather then ones being build via the  debtransform mechanism.

In our usage of OBS we're purely using full source packages rather then debtransform and we want to avoid having spurious source packages in our repositories for  binnmu builds.

Conan-Kudo:
That makes no sense. How else would you verify that the outputs can be fed back as inputs to feed back as outputs?

bluca:
> If the .dsc is what's gets committed, then it's that .dsc (if anything) that should be put through Lintian.
>
> Running Lintian on the output .changes file still analyzes the contents of the binary packages (.deb).

(hi - I think we spoke at FOSDEM last year after the excellent OBS + Debian talk :-) )

In the debtransform case there's no dsc and tarball that get committed. So this would mean missing some important checks that are done on the source pkg, like "source-is-missing", and IMHO it would be a regression.

> What this patch intends to do is to move it to the same model as Debian BinNMUs which is for pure recomplications without source changes in the Debian world (only a package version bump to reflect the rebuild).

Note that OBS can already do this: just add OBS_DCH_RELEASE=1 to your dsc files, and source refresh + rebuild counters will be added to the version. Then you don't need new uploads.

> Well, we want to publish source packages, but we want to publish those we have in OBS already, not their rebuilt versions.

I understand that's one use case. But for example, I have the opposite: I don't care about what's committed, as it gets generated from a git repository and thrown away, I need to publish what can be used to reproduce the binary packages that are also published, which can be modified with build counters (OBS_DCH_RELEASE=1) and so on by the dsc recipe. That's a hard requirement by policy I have, which this change would break.

So, if I may offer an idea for a compromise: what about leaving the default as-is, and adding a server-side config option to skip the source build, that you can enable on your instance? @Conan-Kudo would that be acceptable for you?

Conan-Kudo:
> So, if I may offer an idea for a compromise: what about leaving the default as-is, and adding a server-side config option to skip the source build, that you can enable on your instance? @Conan-Kudo would that be acceptable for you?

I still think the whole idea is fundamentally broken, but as long as it's non-default and people understand what they are giving up in doing so, then fine.

andrewshadura:
> That makes no sense. How else would you verify that the outputs can be fed back as inputs to feed back as outputs?

There’s no such requirement, since the .dsc you upload is the authoritative one. And if you do that, you don’t want the source package to be tainted by the build process, which (unfortunately) may happen with some packages.

Reproducibility is usually a separately achieved goal, and e.g. reproducible-builds.org are only concerned about binary packages being reproducible, not source packages. Building a source package is primarily the maintainer’s job, not the builder’s.

> I still think the whole idea is fundamentally broken, but as long as it's non-default and people understand what they are giving up in doing so, then fine. It should also be per-project/per-package, and not system-wide.

I don’t see why, given .dsc + .debian.tar.?z + .orig.tar.?z is how everyone else is building Debian packages, so logically this workflow should be the default.

bluca:
> I don’t see why, given .dsc + .debian.tar.?z + .orig.tar.?z is how everyone else is building Debian packages, so logically this workflow should be the default.

Well because it would break existing workflow in production deployments, and that's not very nice :-)
Wouldn't it be an acceptable compromise? Instead of patching the source code you can then set an option in your deployment, and existing instances don't get disrupted, with disgruntled users knocking on the door here

andrewshadura:
OBS could automatically detect the workflow used and enable or disable `-b` as needed:

1. any of the debtransform files found ⇒ build both source and binary packages (don’t add `-b`)
2. only `.debian.tar.gz` + `.dsc` found ⇒ build binary packages only, use the source package the maintainer uploaded (add `-b`)

andrewshadura:
Also *I don’t see why* referred to the *fundamentally broken*. I agree that having this as a per-project and per-package option would be great, but I think the default should depend on the type of package, and not be hardcoded as the current default.

sjoerdsimons:
Fwiw; There seems to be the expectation that this patch causes the source package not to be published in the published repositories. That's not the case and not the intention (as that would be very wrong). 

What end up being published is the source package *without* a build counter and the binary builds with build counters, again just like a binnmu in debian. At least if debtransform isn't used.

bluca:
> only .debian.tar.gz + .dsc found ⇒ build binary packages only, use the source package the maintainer uploaded (add -b)

Again, this breaks the existing workflow we use. And given it's the current behaviour, I expect I won't be the only one.

andrewshadura:
>> only .debian.tar.gz + .dsc found ⇒ build binary packages only, use the source package the maintainer uploaded (add -b)
> Again, this breaks the existing workflow we use. And given it's the current behaviour, I expect I won't be the only one.

What workflows would it break? I can’t think of a workflow involving having proper Debian source packages as the source where the current behaviour would be desired or useful. As I see it, only workflows involving debtransform don’t want the proposed behaviour, which will not be affected if the autodetection I’m proposing is implemented, and in the unlikely case someone would need the current behaviour, they would be able to opt out?

bluca:
> What workflows would it break? I can’t think of a workflow involving having proper Debian source packages as the source where the current behaviour would be desired or useful. As I see it, only workflows involving debtransform don’t want the proposed behaviour, which will not be affected if the autodetection I’m proposing is implemented, and in the unlikely case someone would need the current behaviour, they would be able to opt out?

I mentioned it before: in my use case, I don't care about what gets uploaded (which is source packages generated from a git repo on the fly, so no debstransform). But I _need_ the source package that gets published to correspond exactly to the binary packages that get published in the apt repo, including modifications like build counters. It's a hard requirement, by policy. I understand that you have different requirements, and I'm perfectly OK with an option to change this behaviour.
But given that's the current OBS behaviour, I'm sorry, but I'm afraid I need to insist and ask that it doesn't get broken. And @Conan-Kudo has other concerns, but is fine with a disabled-by-default option.

Given you are already going so far as to patching the code right now, isn't it an acceptable compromise to have to set an option to enable the behaviour you want, instead of patching?

Conan-Kudo:
> And @Conan-Kudo has other concerns, but is fine with a disabled-by-default option.

To clarify, I want the existing behavior by default. This new behavior (which I don't like for reasons already outlined), should _never_ be default. It should be able to be enabled on a per package/project level when desired, otherwise never.

adrianschroeter:
closing this for now. Feel free to reopen, if you come up with code which makes this optional. And ideally it would affect then all build types (also no src.rpm for example).

PR #460, opened 2018-08-29
build-recipe-debootstrap: add fallback for Debian SID distro

alee-ccu:
When attempting to build Debian unstable, distribution is
calculated using lsb, which relies on apt cache.

As our build systems lack apt cache and network we cannot detect
we are building for Debian SID (aka unstable).

For more background information see:
  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=845651

Tested-by: Héctor Orón Martínez <hector.oron@collabora.co.uk>

### DIFF
--- a/build-recipe-debootstrap
+++ b/build-recipe-debootstrap
@@ -31,7 +31,11 @@ recipe_prepare_debootstrap() {
 
 recipe_build_debootstrap() {
     local arch=$(chroot $BUILD_ROOT su -c "dpkg-architecture -qDEB_BUILD_ARCH")
-    local dist=$(chroot $BUILD_ROOT su -c "lsb_release --codename --short")
+    if chroot $BUILD_ROOT grep -q '/sid' /etc/debian_version ; then
+        local dist=unstable
+    else
+        local dist=$(chroot $BUILD_ROOT su -c "lsb_release --codename --short")
+    fi
     local myroot=debootstraproot
     test -d $BUILD_ROOT/.build.binaries || cleanup_and_exit 1
     if test "$DO_INIT" = true -o ! -d "$BUILD_ROOT/.build.binaries/dists" ; then


alee-ccu:
Hi. I am `obs-build` package maintainer in Debian:

https://sources.debian.org/src/obs-build/20180302-3/debian/patches/debootstrap-add-fallback-for-Debian-SID-distro.patch/

build-recipe-debootstrap
```diff
@@ -31,7 +31,11 @@ recipe_prepare_debootstrap() {
 
 recipe_build_debootstrap() {
     local arch=$(chroot $BUILD_ROOT su -c "dpkg-architecture -qDEB_BUILD_ARCH")
-    local dist=$(chroot $BUILD_ROOT su -c "lsb_release --codename --short")
+    if $(chroot $BUILD_ROOT grep -q '/sid' /etc/debian_version) ; then
```
adrianschroeter: the $( ) is not needed here, are they?

bluca:
The testing release also will have "/sid" in /etc/debian_version. If you boot a Buster image for example, it's "buster/sid", so won't this be an issue? If you boot a testing image, it will be identified as unstable.

For a similar topic in the deb recipe see: https://github.com/openSUSE/obs-build/pull/307
I raised a ticket for this issue, but it was deemed as working as intended: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=845651

bluca:
In that other case for the deb recipe, we resolved to always identify as the testing release rather than unstable.

build-recipe-debootstrap
alee-ccu: Updated.

smcv:
> In that other case for the deb recipe, we resolved to always identify as the testing release rather than unstable.

For debootstrap, this doesn't really matter: testing is very close to unstable, so the same debootstrap script should be able to bootstrap either one. `unstable` is a bit easier because that script name isn't a moving target.

(In practice all Debian releases since Debian 7 'wheezy' (2013) use a symlink to the `sid`, aka `unstable`, script - only really old releases like squeeze and etch get their own scripts.)

andrewshadura:
Is there anything blocking this from getting merged?

PR #462, opened 2018-08-31
Added policy-rc.d that turns invoke-rc.d into a no-op

alee-ccu:
The obs-build seems work this way:
The "first-stage bootstrap" (OBS unpacking packages which is what this patch fixes) and the "second-stage bootstrap" (which just runs `debootstrap`).

The problem is that in older versions of Ubuntu, the unpack packageset from the target distribution to be able to run the second-stage debootstrap is not very minimal.

In particular, it includes packages like initscripts, udev and procps, which have initscripts that are started by postinst. In older Ubuntu, when a postinst runs invoke-rc.d, it communicates with Upstart via a mechanism that can escape from the chroot and talk to the host system's Upstart (!); and if it can't, invoke-rc.d will fail and the first-stage bootstrap will also fail.

The symptom is that on workers whose host OS did not boot using Upstart (for example Debian stretch, but not Ubuntu trusty), the first-stage bootstrap fails with:

 [   22s] initctl: Unable to connect to Upstart: Failed to connect to
 socket /com/ubuntu/upstart: Connection refused

So here is a patch to provide OBS-level solution that adds a policy-rc.d that turns invoke-rc.d into a no-op, just like debootstrap does to prevent processes in deb build chroots from trying to start system
services.

Signed-off-by: Simon McVittie <smcv@collabora.com>
Tested-by: Héctor Orón Martínez <hector.oron@collabora.co.uk>


adrianschroeter:
can we do this only when upstart is used on the platform at all?

alee-ccu:
The problem is it fails on the host system without upstart, eg: Debian stretch.

alee-ccu:
Solved conflicts.

smcv:
This is functionally equivalent to #484, which was merged, so I think this one can probably be closed?

dbnicholson:
Sorry, I didn't see this when I submitted mine. This looks to be equivalent.

alee-ccu:
Close this one.

> This is functionally equivalent to #484, which was merged, so I think this one can probably be closed?

Sure. Let's close this one.

PR #466, opened 2018-09-24
Use virtio-serial on newer kernels

bmwiedemann:
Assuming that all 4.x target kernels have virtio-serial
This avoids such ugly+noisy messages in output:
  `serial8250: too much work for irq4`

It makes diffing of build logs easier to help reproducible builds.

https://bugzilla.redhat.com/show_bug.cgi?id=986761
https://bugs.launchpad.net/qemu/+bug/1719339

### DIFF
--- a/build-vm-kvm
+++ b/build-vm-kvm
@@ -233,7 +233,11 @@ vm_startup_kvm() {
 	qemu_args=("${qemu_args[@]}" -drive file="$VM_SWAP",format=raw,if=none,id=swap,serial=1,cache=unsafe -device "$kvm_device",drive=swap)
     fi
     # the serial console device needs to be compiled into the target kernel
-    # which is why we can not use virtio-serial on other platforms
+    # which is why we can not use virtio-serial on older platforms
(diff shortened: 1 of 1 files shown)


adrianschroeter:
hm, all kernels since version 4 can have virtio-serial, but there is no guarantee that it is compiled in or that the matching module is available in initrd. Not sure about this...

bmwiedemann:
if the associated kernel config was available somewhere, we could grep there.
```
rpm -qf /boot/config-4.12.14-lp150.12.16-default
kernel-default-4.12.14-lp150.12.16.1.x86_64
```

bmwiedemann:
or maybe we can use `lsinitrd | grep virtio_console.ko` ... will give that a try.

bmwiedemann:
I tried
```bash
if xzcat $vm_initrd | cpio -t | grep -q 'kernel/drivers/char/virtio_console.ko$' ; then
```
but the driver is not part of the initrd. But virtio_serial still works, because the relevant work happens after initrd in the rootfs.

adrianschroeter:
This has been implemented here now:  https://github.com/openSUSE/obs-build/pull/566

PR #467, opened 2018-09-29
Save the _statistics file in the $BUILD_ROOT dir

marcus-h:
Currently, the _statistics file is saved in $TOPDIR/OTHER instead
of $BUILD_ROOT/$TOPDIR/OTHER.

Fixes: https://github.com/openSUSE/osc/issues/449 ("osc build
creates _statistics outside chroot")

### DIFF
--- a/build-recipe
+++ b/build-recipe
@@ -191,7 +191,7 @@ recipe_needs_build_binaries() {
 }
 
 recipe_build_time_statistics() {
-    if test "$DO_STATISTICS" = 1 -a -n "$RECIPE_BUILD_START_TIME" -a -n "$TOPDIR" ; then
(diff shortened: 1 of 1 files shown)


adrianschroeter:
I don't mind to merge it, but _statistics are only written in VM case and not chroot, no?
(in VM case BUILD_ROOT is always just "/")

Or should we maybe disable writing _statistics file in some case?

marcus-h:
On 2018-10-02 02:17:09 -0700, Adrian Schröter wrote:
> I don't mind to merge it, but _statistics are only written in VM case and not chroot, no?
> (in VM case BUILD_ROOT is always just "/")
> 
It is also written in case of a chroot build (cleanup_and_exit calls
recipe_build_time_statistics, which writes the total build time to
the _statistics file). Admittedly, I didn't check the description of
the --statistics option - that's why I thought it might be "useful"
in the non-vm case.

> Or should we maybe disable writing _statistics file in some case?
> 
Yes, I think that's the better approach because it fits to the
description of --statistics. I just forced pushed the corresponding
fix.

PR #469, opened 2018-10-05
Disable dir_index ext2/3/4 option

bmwiedemann:
This reduces indeterminism from filesystem readdir order,
e.g. this snippet suddenly delivers the same output every time:
```bash
mkdir -p in; for i in $(seq 10) ; do touch in/$i ; done
find in/ -type f > unreproducible
```

Since we usually do not have large directories, it should not slow
down the build processes.

Quick testing with
```bash
osc co openSUSE:Factory/python-service_identity && cd $_
time osc build --no-service --clean --vm-type=kvm \
    --offline --build-opt=--vm-disk-filesystem=ext4
```
showed that the difference is 0 +-0.2%

Alternative approach is to use a constant hash_seed via
https://github.com/tytso/e2fsprogs/commit/e1f7100643a46456be107b33098f6034b0835e6d
but that is not yet in Leap 15.0

Helps to limit the effect of problems like
http://bugzilla.opensuse.org/show_bug.cgi?id=1110339

### DIFF
--- a/build
+++ b/build
@@ -332,6 +332,9 @@ Known Parameters:
   --vm-disk-filesystem TYPE
               Defaults for automatic setup of VM root/swap files.
               May get overruled by build config vmfstype build flag.
+  --vm-disk-dirindex
+              On ext2/3/4 enable the dir_index feature.
+              Makes readdir order indeterministic.
 
   --vm-memory SIZEINMB
               Set amount of RAM for VMs
--- a/build-vm
+++ b/build-vm
@@ -22,13 +22,8 @@
 ################################################################
 
 # defaults for vm_img_mkfs
-vm_img_mkfs_ext4_options='-O ^has_journal,^huge_file,^resize_inode,sparse_super'
-vm_img_mkfs_ext4_extra='-E lazy_itable_init,discard'
-vm_img_mkfs_ext4="mkfs.ext4 -m 0 -q -F $vm_img_mkfs_ext4_options"
 vm_img_tunefs_ext4='tune2fs -c 0'
-vm_img_mkfs_ext3='mkfs.ext3 -m 0 -q -F'
 vm_img_tunefs_ext3='tune2fs -c 0 -o journal_data_writeback'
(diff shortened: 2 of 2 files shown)


mlschroe:
Hmm, but is that a good thing? We want to find and fix cases where the build depends on the filesystem ordering, don't we?

bmwiedemann:
We can still find and fix those outside of OBS by having a local revert of this.
I thought, in OBS we want to have reduced rebuilds and re-publishs from indeterminism.

btw: does OBS use ext4 to build?

bmwiedemann:
found it: `/usr/lib/build/queryconfig` reads `.osc/_buildconfig-standard-x86_64` to see
```
# Use ext4 as FS in builder workers (instead of default ext3)
BuildFlags: vmfstype:ext4
```

bmwiedemann:
Maybe I should make it an option, so I can enable/disable it easily for testing if it makes a difference...

bmwiedemann:
dir_index is now an option and default-off here.

mlschroe:
Why is the default off?

bmwiedemann:
default off, to make builds more reproducible by default.
The only reason to switch it on would be performance, if there were cases with 10000+ files in a single dir happening during build.

mlschroe:
I'm not in favor of changing the current default. Lets see what the others have to say.

adrianschroeter:
default change is indeed not a good idea. factory project might decide for it and change it there

adrianschroeter:
@bmwiedemann shall I adapt the code to make it an opt-in?

bmwiedemann:
dir_index is already an opt-in option. Or you want it unreproducible by default with a `--vm-disk-no-dirindex` option?

adrianschroeter:
I have submitted a reworked version now:

https://github.com/openSUSE/obs-build/pull/634

it is really making this an opt-in without changing the default. It can also be configured via prjconf for OBS setups now.
So closing this one.

PR #471, opened 2018-10-15
Allow packets to overwrite the macro _build_insert_debug_package

rhabacker:
Creating mingwxx projects require a different implementation for creating debug packages than native packages. To allow this, the_build_insert_debug_package macro is moved to the macros.build
file, making it overwritable by packages.

I tested this with a local osc installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,8 @@ install:
 	    $(DESTDIR)$(pkglibdir)/Build \
 	    $(DESTDIR)$(pkglibdir)/emulator \
 	    $(DESTDIR)$(bindir) \
-	    $(DESTDIR)$(man1dir)
+	    $(DESTDIR)$(man1dir) \
+	    $(DESTDIR)$(sysconfdir)/rpm
 	install -m755 \
 	    build \
 	    vc \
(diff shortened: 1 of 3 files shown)


mlschroe:
This would only work if the "build" package always gets installed in the build environment, which is not the case. So the code should not package the file, but simply write it like with ~/.rpmmacros.

I'm a bit curious: which what do you overwrite the macro?

rhabacker:
mingw32 and mingw64 packages (https://build.opensuse.org/project/show/windows:mingw:win32 and https://build.opensuse.org/project/show/windows:mingw:win64) have their own process for creating debug packages based on historical decisions, which does not harmonize with the creation of native debug packages. Therefore, the repository option to create debug packages [1] is disabled for these projects as well. 
However, this also means that for a package that contains cross compilers, for example, no native debug packages can be created.
With the patch it would be possible to make this possible again by using a changed variant of %_debug_package for the mentioned project with activation of [1]. 
Do you have a better idea how to solve this issue ?

rhabacker:
> Why not just override the the debuginfo generator macro? 
You are refering to https://fedoraproject.org/wiki/Packaging:Debuginfo#Useless_or_incomplete_debuginfo_packages_due_to_other_reasons

adrianschroeter:
Closing this, because this should be solved without a code change as proposed

PR #484, opened 2019-01-03
deb: Install policy-rc.d to prevent services from starting

dbnicholson:
Per invoke-rc.d(8), a local policy script can be created at
/usr/sbin/policy-rc.d to influence how services are started. When this
script exits with code 101, the requested service is not started. This
is independent of init system and is respected by all maintainer
scripts that comply with debian policy.

This is how debootstrap prevents services from running in the chroot. On
more recent debian and ubuntu releases where systemd is in use, it will
prevent services from starting by detecting that it's in a chroot, but
this method works independently of the init system. In our case this
fixes builds on Ubuntu Trusty where upstart is init. This change likely
makes the hack to replace /sbin/start with a symlink to /sbin/true
unnecessary. That hack is not comprehensive since it only handles the
start action and not stop, restart, reload, etc.

### DIFF
--- a/build-pkg-deb
+++ b/build-pkg-deb
@@ -50,6 +50,16 @@ deb_setup() {
     :>> $BUILD_ROOT/var/lib/dpkg/available
     :>> $BUILD_ROOT/var/log/dpkg.log
     :>> $BUILD_ROOT/etc/ld.so.conf
+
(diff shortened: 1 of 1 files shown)


smcv:
#462 was another submission of essentially the same change, which is obsoleted by this one.

PR #485, opened 2019-01-04
set random.trust_cpu=1 in vm boot cmdline

crrodriguez:
Build hosts should never block getrandom() waiting for entropy
if CPU supports RDRAND/RDSEED.

### DIFF
--- a/build-vm
+++ b/build-vm
@@ -36,7 +36,7 @@ vm_img_mkfs_xfs='mkfs.xfs -f'
 # ignore not backward compatible ext fs options like metadata_csum
 # Only protecting nonroot from root inside guest -> but anyone can be root inside guest
 # so disabling spectre/meltdown mitigations doesn't hurt security and gains performance
-vm_linux_kernel_parameter="ext4.allow_unsupported=1 kpti=off pti=off spectre_v2=off"
+vm_linux_kernel_parameter="ext4.allow_unsupported=1 kpti=off pti=off spectre_v2=off random.trust_cpu=1"
 
 # guest visible devices
 VM_ROOTDEV=/dev/hda1


adrianschroeter:
Not sure if we really want this in the code, because it is cpu specific if one could trust the hardware or not.

Also builds should be reproducible and not contain a secret so a random generator usage sounds not like a good idea (yes, can not get avoided always).

However, what about to add a switch to extend vm_linux_kernel_parameter variable instead as generic solution for individual needs?

crrodriguez:
> Not sure if we really want this in the code, because it is cpu specific if one could trust the hardware or not.

You already trust the cpu...this pertains to the initial seeding of the CSPRNG, once seeded, it will use it the CPU rdrdand/rdseed instruction anyway.

> 
> Also builds should be reproducible and not contain a secret so a random generator usage sounds not like a good idea (yes, can not get avoided always).

It already does, the kernel uses rdrand/rdseed anyways after the initial seeding is performed, unless  nordrand is passed in the command line at boot. what I 'm trying to avoid here is the boot blocking build until enough entropy has been collected from other (hardware) sources (if any). You will start seeing this problem sooner or later, because openssl uses getrandom() *without* RND_NONBLOCK flag. (this is the expected behaviour)

adrianschroeter:
On Dienstag, 12. Februar 2019, 14:06:15 CET Cristian Rodríguez wrote:
> > Not sure if we really want this in the code, because it is cpu specific if
> > one could trust the hardware or not.
> You already trust the cpu...this pertains to the initial seeding of the
> CSPRNG, once seeded, it will use it the CPU rdrdand/rdseed instruction
> anyway.
> > Also builds should be reproducible and not contain a secret so a random
> > generator usage sounds not like a good idea (yes, can not get avoided
> > always).
> It already does, the kernel uses rdrand/rdseed anyways after the initial
> seeding is performed, unless  nordrand is passed in the command line at
> boot. what I 'm trying to avoid here is the boot blocking build until
> enough entropy has been collected from other (hardware) sources (if any).
> You will start seeing this problem sooner or later, because openssl uses
> getrandom() *without* RND_NONBLOCK flag. (this is the expected behaviour)

Is a running haveged helping in your use case?

-- 

Adrian Schroeter
SUSE Linux Products GmbH,  Maxfeldstr. 5, 90409 Nuernberg, Germany
email: adrian@suse.de

crrodriguez:
> Is a running haveged helping in your use case?

haveged comes up later in the boot process, the behaviour you want for build hosts, is getrandom() without  GRND_RANDOM never blocking any process, this includes PID1. (what PID 1 currently does is:

is getrandom() blocking ? 
if yes ..
is this a x86 machine that supports rdrand/rdseed ? 
use that until the kernel stops blocking. 
All of this because old kernels do not have random.trust_cpu=1 or people do not enable this by default.

Other software do not have fallbacks, and will block until initial seeding is done, in virtual machines with no other sources of entropy I have seen it take over 5 seconds to complete, there are reports of it taking  30 seconds or more.

crrodriguez:
The kernel config files were fixed to enable this by default

* Sun Mar 17 2019 mkubecek@suse.cz
- config: enable RANDOM_TRUST_CPU

So this is no longer strictly needed.

PR #488, opened 2019-01-15
dummyhttpserver: also HTML-escape '

maltek:
For now this doesn't really matter since the generated HTML code doesn't use ' for attributes anywhere, but just to be on the safe side in case this changes some time in the future.


adrianschroeter:
well, there would be even more chars what would need escaping. Sorry, I don't see the need for it atm.

PR #492, opened 2019-01-31
RFC: Enable to use this on case insensitive filesystems

hennevogel:
Where Build and build are the same.


hennevogel:
TLDR of the linked OBS issue:

This is the only thing that keeps people from being able to run the OBS development environment on case insensitive file systems. Mostly Mac OS which is hugely popular with Ruby on Rails developers.

marcus-h:
On 2019-01-31 12:14:44 -0800, Henne Vogelsang wrote:
> This is the only thing that keeps people from being able to run the OBS development environment on case insensitive file systems.

Hmm sounds scary - especially in a development env where people might
quickly do something like
- create project "test"
- create another project "Test"
- delete project "test"

which trashes the backend's srcrep, which leads to inconstencies between
the api and the backend etc.

(No opinion about the PR itself...)

hennevogel:
> - create project "test"
> - create another project "Test"

Not really problem as on Mac OS/Windows docker runs inside a Linux VM. It's just that you can not even check out the code base (we use a submodule to this project)  without running into the problem that the files are the same.

Conan-Kudo:
Have you considered renaming the module to `obs-build` instead...? Or even `OBS::Build`?

mlschroe:
I don't think it makes sense to change the layout in the source if the package content stays the same.

hennevogel:
> I don't think it makes sense to change the layout in the source if the package content stays the same.

Yeah sure, this should also be done in the package I guess. I just can not really oversee the consequences. You are the experts! Can we do this? :smile:

hennevogel:
hello?

mlschroe:
Hi there!

hennevogel:
Should I make any changes?

mlschroe:
No.

hennevogel:
This is getting more and more a problem for us during hiring. Can we merge this some time? :-)

adrianschroeter:
We do not see value to introduce regressions just to support building on systems we do not support as target. Closing for now.

Issue #321, opened 2017-01-10
[ArchLinux] GPG verification support


ItachiSan:
`makepkg` allow GPG verification on tarballs.
However, to do so, the related GPG key should be imported in the GPG keyring of the building user.
There is a specific field in the *PKGBUILD* (the `validpgpkeys` array) that contains the fingerprints of the GPG keys needed for verification.
Such fingerprints can be extracted and added to the build OBS keyring before creating the chroot for the package build.

PR #494, opened 2019-03-05
[ArchLinux] Skip unsupported GPG check

jschueller:
GPG signing is not currently supported (https://github.com/openSUSE/obs-build/issues/321)
We want to disable the check so PKGBUILDs relying on `validpgpkeys` can still be built.


jschueller:
When can we expect this to land in the main obs instance ?

PR #497, opened 2019-03-21
Log more information about the build environment and results

smcv:
Debian's sbuild (which is conceptually similar to obs-build, but specific to deb packages) includes a lot of contextual information in its log, which in practice is very useful when reproducing and debugging build failures that happen on remote infrastructure where the package maintainer doesn't necessarily have shell access.

This branch adds:

* The version of obs-build itself, so that developers can tell which features should be available and which bugs have or haven't been fixed
* The exact command used for debootstrap (if used), since it's subtle and can easily go wrong
* The versions of packages that get preinstalled, so that if they fail to unpack or configure, developers can know they are looking at the correct package
* General information about the build chroot, including the kernel, mounted filesystems and a manifest of packages in the chroot

This makes the build log a lot more self-contained when attempting to reproduce and fix build issues.

It also adds details of the packages that were built (in 3b6e6a2), including their metadata (dependencies etc.) and file lists. Again, this mimics what sbuild does. It could be split out into a separate pull request if preferred.


adrianschroeter:
We are a bit unsure here if it would not make sense to have build logs between the different build receipes more similar. On the other hand, I see your point of becoming similar to sbuild ...

nickbroon:
I think this would be very useful. Knowing the debootstrap command and details of the build chroot would have saved me time with analysing builds in the past. sbuild is very useful as debian package tool and would be great example here.

nickbroon:
@smcv @adrianschroeter any prospect of this being merged?
The additional logs would definitely make analysing failed builds a lot easier.

Conan-Kudo:
For what it's worth, Mock does basically the same thing for Fedora. The version of mock, as well as all command invocations for bootstrapping the environment, are included in the logs and output produced by Mock.

andrewshadura:
@adrianschroeter, could you please have a look at this again?

andrewshadura:
As a build and integration engineer for Apertis, I can confirm the information printouts which this MR adds (we run obs-build with these patches applied) has saved me a lot of time spent investigating failed builds. For build dependencies, it helped me reproduce the environment in which the build failed. For outputs, I could immediately see if the packages were missing some files without needing to download them.

Conan-Kudo:
The only problem with this PR is that it doesn't do it for all the build engines `obs-build` supports. At the very least, the RPM and Debian ones need it.

andrewshadura:
Does it have to? The RPM part can be done by someone knowledgeable about RPM (which e.g. I am not, and I think neither is Simon).

Conan-Kudo:
I would be hugely shocked if @smcv didn't know enough about how RPM works to implement this logic for the rpm side, since you're not doing very complicated things in here for the deb side either.

adrianschroeter:
a way to solve this would be to introduce a configuration option (in build config) to enable the more verbose output. It does not need to be implemented in all receipes initially.

PR #502, opened 2019-05-14
add "-vv" parameter in case of debbuild

M0ses:
Without this patch if using the new version (18.12.0) is used all output is
suppressed. This is very confusing as you normally can see the output of
the build process.

This patch enables verbosity level 2, so also the output of the building
process is printed.

This patch shouldn`t harm older versions of debbuild as it had already an
options "-vv". SEE

https://github.com/ascherer/debbuild/blob/18.8.1/debbuild#L304

### DIFF
--- a/build-recipe-spec
+++ b/build-recipe-spec
@@ -130,6 +130,9 @@ recipe_build_spec() {
 	    echo "warning: --nocheck is not supported by this $rpmbuild version"
 	fi
     fi
+    if test "$rpmbuild" == "debbuild" ; then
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
This is not necessary, because I already overrode this with a macro. Please talk to me before you do things like this!

Conan-Kudo:
See: https://build.opensuse.org/package/view_file/Debian:debbuild/debbuild/macros.obs-override?expand=1

By the way, please don't take this as license to update debbuild in `Debian:debbuild`. I'm not done qualifying it locally before updating it for everyone.

PR #503, opened 2019-05-15
Handle cross build

ggardet:
Tested with:
* `osc build aarch64` on `x86_64`
* `osc build x86_64` on `aarch64`

Both are working with `qemu-system-*` without KVM, using kernel/initrd from `kernel-obs-build`

This will fix https://bugzilla.suse.com/show_bug.cgi?id=1135062

### DIFF
--- a/build-vm
+++ b/build-vm
@@ -737,9 +737,14 @@ vm_update_hostarch() {
     local kernel="$vm_kernel"
     local hostarchfile
     local newhostarch
-    if test -z "$VM_KERNEL" -a -e "$BUILD_ROOT/.build.kernel.$VM_TYPE" ; then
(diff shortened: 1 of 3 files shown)


ggardet:
Ping @adrianschroeter

adrianschroeter:
sorry, as written in IRC some time ago, we see the change in KVM type too risky, since it also changes behaviour in native build.

But we want to support this when using --vm-type=qemu , since this mades not much sense anyway anymore. I did start in merging your stuff there .... WIP ...

adrianschroeter:
The core functionality is now in vm_type=qemu implemented here:

 https://github.com/openSUSE/obs-build/pull/534

PR #507, opened 2019-06-20
Adding minimal UTF-8 locales in the chroot

ItachiSan:
The current Archlinux chroot has only the locales:

* C
* POSIX

which makes UTF-8 programs useless (*exempli gratia* the *obs-service-tar_scm* services can't run, due to the missing `en_US.UTF-8` locale).

This commit adds one proper UTF-8 locale, to be correct the `en_US.UTF-8` one.


adrianschroeter:
it does not make sense to do this just for one build format.

This would be also an incompatible change affect builds for old distro versions (if such a thing exists in Arch).

@M0ses IIRC you look for a different solution here in the service, right?

M0ses:
Yes, maybe it`s the best to set LANGUAGE, LANG, LC_ALL to "C.<encoding>" in the obs-service-tar_scm service.

What do you think?

darix:
https://github.com/openSUSE/obs-service-tar_scm/pull/322#issuecomment-514574617

So yes this change could also be the fix. reverting back to C locale for SCM tools is not IMHO.

ItachiSan:
Agree with @darix; I know this is a dirty workaround, but from what I can catch from your comments the other distros have already UTF-8 locales in their chroots/images so yes, this might be a single-distro-only-tweak and is required as more applications relie on UTF-8 (I've struggled quite a lot in the past because of missing locales).

Additionally, as it is a single-distro-tweak, the tweak is within the *build-pkg-arch*; if there is a better function to quickly hack *e.g.* a function running after the *pre-install* and before the *install* step, that should be the place where this tweak should be applied.

adrianschroeter:
let's do in the service package. The build script must be kept forward and backward compatible, so it is definitive not a good place for distro-version specific hacks.

ItachiSan:
@adrianschroeter I guess the OBS chroot is even more broken than I suspected, for not being able to work after the service fix.

PR #508, opened 2019-06-26
configs/15.5.conf: add libzstd1 to preinstall list

DimStar77:


DimStar77:
should address the %check failure seen in https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:A/build/standard/x86_64

PR #516, opened 2019-08-15
sl15.2.conf: add sl15.2.conf for Leap 15.2

nilxam:
I hotpatched this in build package temporary already.
I should removed build service specific stuff in this PR.


adrianschroeter:
sorry, but I need to copy and compare this line by line based on the project config myself. I can not really review a submission from unknown base.

PR #522, opened 2019-09-16
Arch Improvements

Appadeia:
- PKGBUILDS now generate a src.tar.gz package
- arch.conf has a Prefer: iptables to satiate dependency solver


adrianschroeter:
Can you please squash the last two commits?

PR #532, opened 2019-11-11
prepend dash before adding the message

lethliel:
Since we just pass the file with --file=my.changes to
vc and not only the content, vc will not prepend the
"-" anymore.

This will fix https://github.com/openSUSE/osc/issues/670

### DIFF
--- a/vc
+++ b/vc
@@ -136,7 +136,10 @@ set +e
 	if [ -n "$message" ]; then
 		echo -e "- $message"
 	elif [ -n "$content" ]; then
-		cat "$content"
+	        cat "$content" > format_tmp
+                sed -i '1s;^;- ;' format_tmp
+                cat format_tmp
+                rm format_tmp
 	elif [ ! $just_edit ]; then
 		echo "- "
 	fi


krop:
So, noone to accept this PR?

vc
```diff
@@ -136,6 +136,10 @@ set +e
 	if [ -n "$message" ]; then
 		echo -e "- $message"
 	elif [ -n "$content" ]; then
+	        cat "$content" > format_tmp
+                sed -i '1s;^;- ;' format_tmp
+                cat format_tmp
+                rm format_tmp
```
adrianschroeter: these 4 lines can be replaced with

sed -i '1s;^;- ;' "$content"

no?

lethliel: No. Because this way the file the content comes from gets modified. And on a second run there are two dashes. 

And I don't want to modify the files the user gives as source.

lethliel:
@adrianschroeter The detour via the tmpfile is needed. Otherwise the original source content is modified. 

But I removed a superfluous echo statement.

lethliel:
not needed as discussed in https://github.com/openSUSE/osc/issues/670

PR #533, opened 2019-11-12
Add systemd-nspawn backend

OlegGirko:
This change adds support for building inside a lightweight containers created by `systemd-nspawn`.

As Fedora 31 switched to cgroup v2, `lxc` can't support limiting device access to host system anymore.

There are no files in cgroup v2 hierarchy that provide interface to rules for device access: eBPF program should be used instead.

Also, Docker does not work as well with cgroup v2.

This leaves us without working backends supporting lightweight containers, so a new backend that can work with cgroup v2 becomes necessary.

### DIFF
--- a/build-vm-nspawn
+++ b/build-vm-nspawn
@@ -1,0 +1,73 @@
+#
+# systemd-nspawn specific functions
+#
+################################################################
+#
+# Copyright (c) 2019 Oleg Girko
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


build-vm-nspawn
```diff
@@ -0,0 +1,74 @@
+#
+# Docker specific functions
```
Conan-Kudo: This should be `nspawn`

build-vm-nspawn
```diff
@@ -0,0 +1,74 @@
...
+#
+################################################################
+#
+# Copyright (c) 2019 Oleg Girko
+# Copyright (c) 2019 SUSE Linux Products GmbH
```
Conan-Kudo: Don't put SUSE copyright unless there's a reason to...

build-vm-nspawn
OlegGirko: Oops, missed this one. I was using `build-vm-docker` as a template. Fixed.

build-vm-nspawn
OlegGirko: OK, removed.

mlschroe:
Thanks!

PR #536, opened 2019-12-10
do not retry automatically on disk full

adrianschroeter:
we have constraints meanwhile in OBS


adrianschroeter:
Let's discuss tomorrow if we can get rid of this really ...

adrianschroeter:
will adapt to do it only in non-VM mode

PR #537, opened 2019-12-13
allow to enter an existing buildroot w/o recipe

lethliel:
The chroot needs to be done before the recipe code.
Otherwise one can not enter a buildroot from outside
a working directory.

In addition a recipe is not needed to chroot into
an existing buildroot.

Is needed in addition to https://github.com/openSUSE/osc/pull/710

### DIFF
--- a/build
+++ b/build
@@ -1131,6 +1131,12 @@ fi
 
 set_build_arch
 
+if test "$DO_INIT" = false -a -n "$RUN_SHELL"; then
+    mount_stuff
+    chroot $BUILD_ROOT su - $BUILD_USER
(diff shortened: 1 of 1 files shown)


lethliel:
@mlschroe please have a look

lethliel:
@mlschroe In addition with https://github.com/openSUSE/osc/pull/710 this will restore the old behavior of osc chroot / osc shell. 

The next step would be to implement the build-shell argument. But this will be done in a separate PR. This one is just for fixing.

lethliel:
as discussed mit @mlschroe this can be merged

PR #538, opened 2020-01-03
configs: fix testbuild fail with latest rpm in SP2 and Leap 15.2

nilxam:
Since latest rpm in SP2 has BuildRequires libgcrypt-devel to use libgcrypt as crypto library, build config need to adjusted to add preinstall for crypto library.

build fail link: https://build.opensuse.org/package/live_build_log/openSUSE:Leap:15.2:Staging:A/build/standard/x86_64
rpm update: https://build.opensuse.org/request/show/760171

### DIFF
--- a/configs/sl15.2.conf
+++ b/configs/sl15.2.conf
@@ -51,6 +51,9 @@ Prefer: libffi7
 
 Preinstall: liblua5_3-5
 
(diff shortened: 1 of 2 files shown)


lnussel:
I suppose the same change is required for Factory?

PR #540, opened 2020-01-09
use virt-make-fs when using ext3 base file system

sjamgade:
the patch add functionality to use a folder (BUILD_ROOT) as base image(VM_ROOT) for kvm based builds by creating the image after BUILD_ROOT is populated there is no `mount` call required. This is necessary for kvm based build in containerized environments with access to `/dev/kvm`. 

The patch checks the `virt-make-fs` binary is available and vmdisk-filesystem is ext3 (restricted as of now).  THe patch also splits `vm_setup()` function into two parts (setup_swap, and setup root) to highlight the difference between the two setups. Swap setup is kept as was. While root (VM_ROOT) setup is conditioned to (presence of virt-make-fs binary and ext3 filesystem).  So the feature can be completely disabled if the binary is not available.

### DIFF
--- a/build-vm
+++ b/build-vm
@@ -662,26 +662,63 @@ vm_set_defaults() {
 }
 
 #
-# create file system and swap space, mount file system to $BUILD_ROOT
+# create swap space
+#
+vm_setup_swap() {
+    if test -n "$VMDISK_CLEAN" ; then
+	# delete old swap to get rid of the old blocks
+	if test -n "$VM_SWAP" -a "$VM_SWAP_TYPE" = file -a -f "$VM_SWAP" ; then
+		echo "Deleting old $VM_SWAP"
+        rm -f "$VM_SWAP"
+	fi
+    fi
+    if test -n "$VM_SWAP" -a "$VM_SWAP_TYPE" = file ; then
+	vm_img_create "$VM_SWAP" "$VMDISK_SWAPSIZE"
+    fi
+    if test -n "$VM_SWAP" ; then
+	vm_attach_swap
+	dd if=/dev/zero of="$VM_SWAP" bs=1024 count=1 conv=notrunc 2>/dev/null
+	if test "$VM_SWAPDEV" != "${VM_SWAPDEV#LABEL=}"; then
+	    # call mkswap to set a label
+	    mkswap -L "${VM_SWAPDEV#LABEL=}" "$VM_SWAP"
+	fi
+	vm_detach_swap
+        # mkswap happens inside of the vm
+    fi
+}
+
+#
+# create file system, mount file system to $BUILD_ROOT
 #
 vm_setup() {
     vm_set_filesystem_type
     vm_set_mount_options
     echo "VM_ROOT: $VM_ROOT, VM_SWAP: $VM_SWAP"
 
+    USE_VIRT_FS=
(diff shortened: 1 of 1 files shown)


build
```diff
@@ -1155,7 +1155,11 @@ if test -n "$LIST_STATE" ; then
 fi
 
 # do vm setup if needed
+USE_VIRT_FS=
```
mlschroe: I think it's cleaner to do this in vm_setup and leave this file unmodifired

build-vm
```diff
@@ -662,66 +662,20 @@ vm_set_defaults() {
 }
 
 #
-# create file system and swap space, mount file system to $BUILD_ROOT
+# create swap space
 #
-vm_setup() {
-    vm_set_filesystem_type
-    vm_set_mount_options
-    echo "VM_ROOT: $VM_ROOT, VM_SWAP: $VM_SWAP"
```
mlschroe: I think we should leave the echo call as it is instead of splitting it into two calls

build-vm
```diff
@@ -662,66 +662,20 @@ vm_set_defaults() {
...
-
-    vm_attach_root
-    # this should not be needed, but sometimes a xen instance got lost
-    test "$VM_TYPE" = xen && vm_purge_xen
+setup_swap() {
```
mlschroe: Please rename to vm_setup_swap()

build-vm
```diff
@@ -734,6 +688,69 @@ vm_setup() {
...
+	# delete old root to get rid of the old blocks
+	if test -n "$VM_ROOT" -a "$VM_ROOT_TYPE" = file -a -f "$VM_ROOT" ; then
+	    echo "Deleting old $VM_ROOT"
+	    rm -f "$VM_ROOT"
+	fi
+    fi
+    # using virt-make-fs for ext3 file VMDISK_FILESYSTEM
+    # so nothing is needed to do aprt from creating appropriate folders
+    # and calling virt-make-fs accordingly
+    if test ! "$USE_VIRT_FS" ; then
```
mlschroe: Hmm, as the `if` statement basically goes to the end of the function, how about doing
```
if test -n "$USE_VIRT_FS" ; then
    return 0
fi
```

This should make the commit much smaller and less error prone. For example, you've moved the mkdir_build_root line in your commit by mistake.

build-vm
```diff
@@ -915,6 +934,9 @@ vm_first_stage() {
 	    rmdir "$KERNEL_TEMP_DIR"
 	fi
     fi
+    if test "$USE_VIRT_FS" ; then 
+        /usr/bin/virt-make-fs --format=raw -t ext3 --size="$VMDISK_ROOTSIZE"M $BUILD_ROOT $VM_ROOT
```
mlschroe: Just a random thought: shouldn't we clean $BUILD_ROOT after the virt-make-fs call?

sjamgade: We could but we dont have to because there is `mkdir_build_root` call after `vm_setup` which should effectively cleanup `$BUILD_ROOT` before every build, which is the reason I left it.

mlschroe:
There's still lots of unneeded white-space changes, can you please revert that?
There's also the moved mkdir_build_root which is still in the pull request.

sjamgade:
Okay that should fix it

mlschroe:
Still lots of white space changes ;)

Just look at the git "files changed" tab, there's a toggle where you can make the diff ignore white space changes.

mlschroe:
Thanks a lot!

PR #543, opened 2020-01-22
use abuild user if $NOROOTFORBUILD is set.

lethliel:
see $subject


lethliel:
@adrianschroeter please review

PR #545, opened 2020-01-27
Factory preinstall using dep expansion

adrianschroeter:
This is independend if the decision what we use in OBS.


mlschroe:
Wait wait wait, this has to follow the openSUSE:Factory config. And you'll want to minimize the expanded preinstall list first.

adrianschroeter:
IMHO it does not have to follow, since factory is anyway a moving target.

mlschroe:
What's the point of this if it does not match? If you want this, please rename the config to "adrian.conf" or something like that.

adrianschroeter:
the chances are higher that it matches with that change.... as the package lifes also in non-factory distros with that config file.

adrianschroeter:
michael gave up on me:)

PR #547, opened 2020-02-04
call obsgendiff if configured

adrianschroeter:
It should fail on purpose, when no script is available, but the feature
is configured

Not using kiwi_post_run to allow this feature independend of the build
tool


adrianschroeter:
still not happy about "obsgendiff" name .... what about "diff2release" or "releasediffer" ?

lethliel:
LGTM

nickbroon:
What is the purpose of this, and can you give an example of it's use?
(and/or point to the source of /usr/lib/build/obsgendiff ?)

PR #554, opened 2020-03-19
set buildflavor also for Build::parse

adrianschroeter:


substitutedeps
```diff
@@ -95,6 +95,7 @@ my $newspec = $ARGV[1];
 
 my $cf = Build::read_config_dist($dist, $archs, $configdir);
 $cf->{'warnings'} = 1;
+$cf->{'buildflavor'} = $buildflavor if $buildflavor;
```
mlschroe: all other places do `if defined $buildflavor`, we should do the same here.

PR #556, opened 2020-03-26
disable transparent_hugepage on s390x guests for now, causes hangs

bugfinder:


adrianschroeter:
do you have some bugreport where the issue is described?

bugfinder:
https://bugzilla.suse.com/show_bug.cgi?id=1163684

bugfinder:
maybe also https://bugzilla.suse.com/show_bug.cgi?id=1130544, unclear yet

PR #557, opened 2020-03-29
Run /sbin/ldconfig before booting VM

matwey:
In some rare cases, `bash` depends on `libtermcap` package where appropriate
symlinks are missed. In such cases the build script has no chance to start in
VM at the second stage since `/bin/bash` is not functional yet:

```
    /bin/bash: error while loading shared libraries: libtermcap.so.2: cannot open shared object file: No such file or directory
```

Run `/sbin/ldconfig` for preinstalled packages to recover shared object
resolution.


mlschroe:
I'm sorry, we cannot merge that. This change would make it trivial to get root access on the worker because one can replace /sbin/ldconfig with a malicious binary. Breaking out of a chroot is trivial.

This really needs to be fixed in the packages, the so symlink needs to be present.

PR #558, opened 2020-04-01
add ability to store ccache when PKG_CCACHE variable is true

sjamgade:
this feature will allow usage of ccache wihtin th build env.
the storage dir is /.ccache. the permission for the $BUILD_USER are
already available.

The activation of feature is dependent on param `--prpa-ccache` being
passed.

The created tar.gz is sent back only when there is a successful build
(non error, non unchanged).

This is the first part where worker only send the generate cache. The
second part where worker can download this cache is being implemetned.

### DIFF
--- a/build
+++ b/build
@@ -106,6 +106,7 @@ DO_STATISTICS=
 RUN_SHELL=
 RUN_SHELL_CMD=
 CCACHE=
+PKG_CCACHE=
 DLNOSIGNATURE=
 BUILD_FLAVOR=
 OBS_PACKAGE=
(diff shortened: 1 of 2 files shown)


mlschroe:
Please change the name, the build tool does not know about a "PRPA".
(I'm also not happy about using a "dot" file for the ccache artifact tar file)

sjamgade:
Yes, I was also unsure about what should be the naming convention. what would you recommend
(`--package-ccache` ,`_ccache.tar.gz`) or (`--download-ccache`, `_{prev,build,built}ccache.tar.gz`) or ....

sjamgade:
#559 - merged

PR #559, opened 2020-04-03
extend --ccache to generate _ccache.tar.gz and implement --pkg-ccache

sjamgade:
with `--ccache` every build will generate a `OTHER/_ccache.tar.gz` archive
which can be used for subsequent builds. However for local builds
explicit path to archive with `--pkg-ccahe` is necessary.

A user provided archive can be used as ccache contents with
`--pkg-ccache=/path/to/archive.tar.gz`

`--pkg-ccache` will override the default ccache. (`--pkg-ccahe` implies
`--ccache` but with provided archive)

### DIFF
--- a/build
+++ b/build
@@ -106,6 +106,7 @@ DO_STATISTICS=
 RUN_SHELL=
 RUN_SHELL_CMD=
 CCACHE=
+PKG_CCACHE=
 DLNOSIGNATURE=
 BUILD_FLAVOR=
 OBS_PACKAGE=
@@ -253,6 +254,10 @@ Known Parameters:
 
   --ccache
               Use ccache to speed up rebuilds
+
+  --pkg-ccache /path/to/ccache.tar.gz
+              path to archive of ccache directory content.
+              Its contents will be extracted to /.ccache
 
   --icecream N
               Use N parallel build jobs with icecream
@@ -490,6 +495,8 @@ setupccache() {
 	chown -R "$ABUILD_UID:$ABUILD_GID" "$BUILD_ROOT/.ccache"
 	echo "export CCACHE_DIR=/.ccache" > "$BUILD_ROOT"/etc/profile.d/build_ccache.sh
 	echo 'export PATH=/var/lib/build/ccache/bin:$PATH' >> "$BUILD_ROOT"/etc/profile.d/build_ccache.sh
+    local ccachetar="$BUILD_ROOT/.build.oldpackages/_ccache.tar.gz"
+    test -e $ccachetar && tar -zxf $ccachetar -C "$BUILD_ROOT/.ccache/"
     else
 	rm -f "$BUILD_ROOT"/var/lib/build/ccache/bin/{gcc,g++,cc,c++,clang,clang++}
     fi
@@ -964,7 +971,13 @@ while test -n "$1"; do
 	test "$icecream" -gt 0 && BUILD_JOBS="$ARG"
 	shift
       ;;
-      -ccache)
+      -ccache) # this is initialized after all the parsing is done
+	CCACHE=true
+      ;;
+      -pkg-ccache)
+	needarg
+	PKG_CCACHE="$ARG"
+	shift
 	CCACHE=true
       ;;
       -statistics)
@@ -1304,6 +1317,9 @@ for RECIPEFILE in "${RECIPEFILES[@]}" ; do
 	    fi
 	fi
 	copy_oldpackages
+        # chroot based builds
+        # rsync as source and dest could be same
+        test -n "$PKG_CCACHE" -a -e "$PKG_CCACHE" && rsync -v "$PKG_CCACHE" "$BUILD_ROOT/.build.oldpackages/_ccache.tar.gz"
     fi
 
     mount_stuff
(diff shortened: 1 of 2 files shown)


build
```diff
@@ -490,6 +495,7 @@ setupccache() {
 	chown -R "$ABUILD_UID:$ABUILD_GID" "$BUILD_ROOT/.ccache"
 	echo "export CCACHE_DIR=/.ccache" > "$BUILD_ROOT"/etc/profile.d/build_ccache.sh
 	echo 'export PATH=/var/lib/build/ccache/bin:$PATH' >> "$BUILD_ROOT"/etc/profile.d/build_ccache.sh
+	tar -zxf "$BUILD_ROOT/.build.oldpackages/_ccache.tar.gz" -C /.ccache/
```
adrianschroeter, line 498: don't we need an existens check here?

(We might do not have a former build result or the method is used in old scenario)

sjamgade: the deliberately left it without check: 

- tar will just throw an error into the log file, indicating an error condition ($ccache is set but the archive is not there). There is no sure shot way to know if this the first build with --ccache (in which case, ofcouse there is no archive) but subsequent builds should have this archive present and if this archive is absent that is an error condition.

I think its an overkill to fail the entire build just because ccache is no present.

sjamgade: Also if the path with  --pkg-ccahe is not correct (typo or any other reason), tar will indicate that archive is not present. As will also be indicated by rsync

adrianschroeter:
On Freitag, 3. April 2020, 14:49:43 CEST wrote Sumit Jamgade:
> @sjamgade commented on this pull request.
> 
> 
> 
> > @@ -490,6 +495,7 @@ setupccache() {
>  	chown -R "$ABUILD_UID:$ABUILD_GID" "$BUILD_ROOT/.ccache"
>  	echo "export CCACHE_DIR=/.ccache" > "$BUILD_ROOT"/etc/profile.d/build_ccache.sh
>  	echo 'export PATH=/var/lib/build/ccache/bin:$PATH' >> "$BUILD_ROOT"/etc/profile.d/build_ccache.sh
> +	tar -zxf "$BUILD_ROOT/.build.oldpackages/_ccache.tar.gz" -C /.ccache/
> 
> the deliberately left it without check: 
> 
> - tar will just throw an error into the log file, indicating an error condition ($ccache is set but the archive is not there). There is no sure shot way to know if this the first build with --ccache (in which case, ofcouse there is no archive) but subsequent builds should have this archive present and if this archive is absent that is an error condition.
> 
> I think its an overkill to fail the entire build just because ccache is no present.

right, I thought more about not to call tar if the file is not there.
And not report an error.

-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

build
```diff
@@ -1272,6 +1287,10 @@ for RECIPEFILE in "${RECIPEFILES[@]}" ; do
     if test -n "$VM_TYPE" -a -z "$RUNNING_IN_VM" ; then
 	vm_first_stage
 	cleanup_and_exit
+    else
```
mlschroe, line 1290: No need for the "else", cleanup_and_exit exits

build
```diff
@@ -1089,6 +1101,9 @@ validate_buildroot "$BUILD_ROOT"
 # done option parsing
 BUILD_OPTIONS_PARSED=true
 
+test -n "$CCACHE" && CCACHE="$BUILD_ROOT/.build.oldpackages/_ccache.tar.gz"
```
mlschroe, line 1104: Why is this ".build.oldpackages" suddenly the default for --ccache?
Why use ".build.oldpackages" at all?

build
```diff
@@ -1272,6 +1287,10 @@ for RECIPEFILE in "${RECIPEFILES[@]}" ; do
     if test -n "$VM_TYPE" -a -z "$RUNNING_IN_VM" ; then
 	vm_first_stage
 	cleanup_and_exit
+    else
+        # chroot based builds
+        # rsync as source and dest could be same
+        test -n "$CCACHE" && rsync -v "$CCACHE" "$BUILD_ROOT/.build.oldpackages/_ccache.tar.gz"
```
mlschroe, line 1293: Hmm, doesn't copy_oldpackages delete this if --oldpackages is not used?

mlschroe:
I don't understand why this is tied to the oldpackages mechanism. We need some special mechanism in the worker/repserver anyway, because

- we do not want 'osc getbinaries' to download the ccache
- I'm partial about showing the ccache as build artifact
- we still need to download the ccache even if the worker does not download the old build result (this happens if the 'nounchanged' flag is present in the build job, i.e. the binary diff mechanism is disabled)

build
sjamgade: yes, I will move the rsync call below the copy_oldpackage call

build
sjamgade: because the ccache tar is from preivously built package.

build
sjamgade: yup, agreed

build
mlschroe: No it's not, it's from the --pkg-ccache argument.

PR #560, opened 2020-04-20
enable compression on ccache

sjamgade:
This allows better storage with slight overhead. Also reduces the
final archive size.

### DIFF
--- a/build
+++ b/build
@@ -505,8 +505,8 @@ setupccache() {
         else
             echo "compression = true" >> $cconf
         fi
-    local ccachetar="$BUILD_ROOT/.build.oldpackages/_ccache.tar.gz"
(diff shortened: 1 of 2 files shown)


mlschroe:
So ccache compresses the artifacts and we then compress the ccache tar ball again? Does the second compression make sense in that case?

sjamgade:
ccache will already have compressed the object files so tar command will have less work to do. Also ccache will to do the inital decompressed only when there is a "hit". So initial tar extract time will also be reduced. 

ccache does object file level compression where as tar will do compression across files. So there is some benefit in having both of them.

mlschroe:
But how much is it you gain with the second compression step? I.e. could you please uncompress the tar and check how much the size changes?

sjamgade:
The second compression was 65% faster and on disk cache size was 50% less. However  the second compression does not have any impact ( tried it on evolution package)  space savings

PR #563, opened 2020-05-06
Add support for Arch in build-recipe-kiwi

davidcassany:
This commit adds support for Arch Linux repositories in the KIWI
build recipe.

### DIFF
--- a/build-recipe-kiwi
+++ b/build-recipe-kiwi
@@ -767,18 +767,21 @@ recipe_build_kiwi() {
 	    repo="$TOPDIR/SOURCES/repos/${rc%/*}/${rc##*/}/"
         fi
         if test "$imagetype" != product -a "$DO_INIT" != "false" ; then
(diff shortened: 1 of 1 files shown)


davidcassany:
I could not fully verify this PR in my host as I realized that `osc` tool perpares repositories in a different way comparted to remote builds. The generated paths for the buildtime repositories do not include the architecture subfolders in remote builds, but they do in local builds when using `osc build` (chroot build). In fact an equivalent code of this PR has been tested [here](https://build.opensuse.org/package/show/home:dcassany:Arch/obs-service-arch_repos_builder) as a buildtime obs service. This PR is done with the assumption that `build-recipe-kiwi` runs in an equivalent environment of what a buildtime OBS service would do.

davidcassany:
@mlschroe thank much for the review. Is there any expectation on when it will be available in the OBS instance?

mlschroe:
I think it's already deployed.

PR #564, opened 2020-05-12
Update for zVM to make container builds work.

azouhr:
- Use ZVM_VOLUME_ROOT and ZVM_VOLUME_SWAP variables to define the local
  zVM environment
- Increase wait to connect iucv console to 5 seconds. This is not
  critical, since the messages are buffered, however I encountered
  connection fails with 2 seconds.
- Add new function vm_initrd_obs_modules_zvm to create new worker initrd
  on demand. This uses the initrd created by the obsworker script and
  updates the kernel modules within according to the OBS kernel.
  The resulting initrd is stored at the same location as the obsworker
  initrd, however expanded with the actual kernel version and thus
  cached for subsequent builds.

### DIFF
--- a/build-vm-zvm
+++ b/build-vm-zvm
@@ -21,17 +21,18 @@
 #
 ################################################################
 
+ZVM_CLEANUP=
+ZVM_VOLUME_ROOT=${ZVM_VOLUME_ROOT:-0100}
+ZVM_VOLUME_SWAP=${ZVM_VOLUME_SWAP:-0200}
+
 # z/VM: use default kernel image from local machine
 # lets go with the default parameters. However zvm_initrd will be a required parameter
-#zvm_kernel=/boot/image
+# zvm_kernel=/boot/image
 #zvm_initrd=/boot/initrd_worker
-zvm_param="root=/dev/disk/by-path/ccw-0.0.0150-part1 hvc_iucv=8 console=hvc0 $vm_linux_kernel_parameter"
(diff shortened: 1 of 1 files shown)


build-vm-zvm
```diff
@@ -21,17 +21,18 @@
 #
 ################################################################
 
+ZVM_CLEANUP=
+ZVM_VOLUME_ROOT=${ZVM_VOLUME_ROOT:-0100}
+ZVM_VOLUME_SWAP=${ZVM_VOLUME_SWAP:-0200}
+
 # z/VM: use default kernel image from local machine
```
bugfinder, line 28: actually why 0100 and not 0150, wasn't 0150 the default root dev ?

build-vm-zvm
```diff
@@ -223,7 +227,6 @@ zvm_worker_init() {
 }
 
 zvm_cp() {
-    modprobe vmcp || zvm_fatal "Cannod load vmcp module"
     if test -n "$1" ; then
```
bugfinder, line 230: is manually loading vmcp no longer required at all ?

build-vm-zvm
azouhr: Good question. we cn change it back, I just found that most people do not use 150 but 100 as base address

build-vm-zvm
azouhr: Its actually not possible anymore with current kernel versions. Its not a module anymore, but compiled into the kernel. The reason why it is not in here at all is, that the obsworker script already tries to load this module (for older distributions) and loading the module repeatedly is not necessary at all.

bugfinder: okay, that answered my question ... looks good to me then

PR #566, opened 2020-05-14
Set kvm_serial_device to virtio-serial in the fixup

mlschroe:


adrianschroeter:
core part works, will follow up with my other stuff in another request

PR #568, opened 2020-05-14
include --noorder option to rpm install when accumulation is enabled …

koomie:
…in order

to preserve project Order declarations.

Signed-off-by: Karl W. Schulz <karl@oden.utexas.edu>

### DIFF
--- a/build-pkg-rpm
+++ b/build-pkg-rpm
@@ -135,8 +135,9 @@ pkg_finalize_rpm() {
 	    rm -f $BUILD_ROOT/${CUMULATED_LIST[$num]}
 	    cp $BUILD_ROOT/.init_b_cache/rpms/$PKG $BUILD_ROOT/${CUMULATED_LIST[$num]} || cleanup_and_exit 1
 	done > $BUILD_ROOT/.init_b_cache/manifest
-	( cd $BUILD_ROOT && chroot $BUILD_ROOT rpm --ignorearch --nodeps -Uh --oldpackage --ignoresize --verbose $RPMCHECKOPTS \
-		$ADDITIONAL_PARAMS .init_b_cache/manifest 2>&1 || touch $BUILD_ROOT/exit )
+	( cd $BUILD_ROOT && chroot $BUILD_ROOT rpm --ignorearch --noorder --nodeps -Uh --oldpackage --ignoresize --verbose \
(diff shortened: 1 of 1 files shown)


koomie:
Motivation here is to honor project configuration Order stanzas when `DO_CUMULATE=true`. Otherwise, the multi-package rpm install will reorder the install package ordering (at least, this is observed with Leap 15.1).

adrianschroeter:
Esp. for SUSE distros this should be fixed in the distro instead.

It makes the build incompatible btw, so we could only introduce this only via a flag.

But it is a bit questionable if we really need this (in first place it would be a workaround for broken packages)

koomie:
Agree that this is not an issue with the distro packages. However, the cumulate option is enforced for all packages necessary for a particular build.  In OpenHPC's case - we pull in some 3rd party packages we don't control (vendor compilers) that don't have their dependencies defined 100% correctly. To resolve, we have historically taken advantage of the [`Order`](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.prjconfig.html#sec.prjconfig.syntax) option in an OBS project config, but the problem with the current code is that package Order specifications are **silently ignored** when cumulate is triggered. It looks to me like the RPM package list is initially ordered correctly to honor any `Order` specifications and that's why I suggested a simple fix with the `--norder` addition.  

Maybe another possibility for those who rely on the `Order` option is to provide an option to the project admin to disable cumulate. Right now it is hard coded via: 
```    DO_CUMULATE=
    typeset -ri suse_version=$(chroot $BUILD_ROOT rpm --eval '%{?suse_version}' 2>/dev/null)
    if ((suse_version > 1220)) ; then
        DO_CUMULATE=true
    fi
```


Or, perhaps there is no interest in maintaining support for the `Order` option with newer suse distro's - in that case perhaps updating the documentation to say that it does not work with `suse_version > 1220` would be helpful? (it would have helped clue me in to why our dependency chains were broken on Leap but not CentOS). 

Thanks for the consideration.

mlschroe:
It does work with newer suse distros. It was added because some distros had *missing* dependencies, it works in that case even without --noorder because rpm does not touch the command line order if there is no dependency.

You seem to need another use case, because you want to get rid of some existing dependency. I'm curious why it is there, and why it is wrong.

If you absolutely need this functionality we can add a BuildFlag: rpmnoorder to enable --noorder installs. Cumulating is still needed for some packages like tetex that rely on the posttrans script being called after all tetex packages are installed.

PR #573, opened 2020-06-05
also package pkg-config files by default into baselibs. (bsc#1172563)

msmeissn:
pkg-config files are interesting for development, so include them here.

e.g. used by wine development of 32bit


mlschroe:
The 32bit packages generally will not work without their 64bit counterpart.

PR #579, opened 2020-06-29
Update build-vm to work with z/VM

azouhr:
z/VM has 4-digit hex numbers for devices.
- update build.data with this number
- map VM_SWAP to a real block device during build
- don't rely on non-persistent devices staying identical

### DIFF
--- a/obs-docker-support
+++ b/obs-docker-support
@@ -314,7 +314,11 @@ EOF
 }
 
 if test `id -u` != "0"; then
-    echo "obs-docker-support: not executed as root user!"
+    echo <<EOF
+obs-docker-support: not executed as root user!
+
+This can happen if your base container (see FROM line) is setting USER already, then builds in dependent container run as that user/group.
+EOF
     exit 1
 fi
 


build-vm
```diff
@@ -566,6 +566,14 @@ vm_detect_2nd_stage() {
 	    echo "resolved swap device $VM_SWAP to $i"
 	    VM_SWAP=$i
 	fi
+        # for zvm, there is just a 4-digig hex number
+            if (( 16#$VM_SWAP )); then
+                if test ${#VM_SWAP} -eq 4; then
```
adrianschroeter, line 571: this looks not like valid bash syntax.

Also some quotes to protect against invalid data seems to be missing

build-vm
```diff
@@ -1024,7 +1039,8 @@ vm_save_statistics() {
     test -e /dev/sda && device="sda"
     test -e /dev/vda && device="vda"
     test -e /dev/xvda && device="xvda"   # in newer XEN setups
-    test -e /dev/dasda && device="dasda" # in z/VM
+    # while there is dasda in z/VM, it can happen that dasda really is the swap device.
+    test -e /dev/dasda && device=$(ls /sys/bus/ccw/devices/0.0.${ZVM_VOLUME_ROOT}/block) # in z/VM
```
adrianschroeter, line 1043: why is the ls command needed here at all?

build-vm
```diff
@@ -566,6 +566,14 @@ vm_detect_2nd_stage() {
 	    echo "resolved swap device $VM_SWAP to $i"
 	    VM_SWAP=$i
 	fi
+        # for zvm, there is just a 4-digig hex number
+            if (( 16#$VM_SWAP )); then
+                if test ${#VM_SWAP} -eq 4; then
+                    sleep 1
+                    VM_SWAP="/dev/$(ls /sys/bus/ccw/devices/0.0.${VM_SWAP}/block/)1"
```
adrianschroeter, line 573: why is the ls command needed here at all?

build-vm
azouhr: hmm ... I actuallyl thought it might be nice to redirect the output to /dev/null, however the code works. First test is, if $VM_SWAP is a hex number, second is, if it has 4 digits:
> (( 16#1f )); echo $?
0
> (( 16#1g )); echo $?
bash: ((: 16#1g: value too great for base (error token is "16#1g")
1

> VM_SWAP=12345
> echo ${#VM_SWAP}
5

build-vm
azouhr: Unfortunately, the build worker skript is not clean with regards to VM_SWAP and VM_SWAPDEV. Therefore, I just set it at that point to the real device name as found in /sys.

build-vm
azouhr: When booting the worker guest, /dev/dasda or /dev/dasdb are not considered persitent devices. To have some kind of persistency, I use the (known) value of the SWAP device to determine the actual device name. Typically will also resolve to dasda, because it has the lower number.

PR #583, opened 2020-07-20
fix dependencies for Fedora 33

M0ses:
With Fedora 33 the POSIX perl module was split out of the perl package.

This patch addeds Requires and BuildRequires for the perl POSIX module.

### DIFF
--- a/dist/build.spec
+++ b/dist/build.spec
@@ -88,6 +88,12 @@ Requires:       %{__pkg_name}-mkbaselibs
 Recommends:     %{__pkg_name}-mkdrpms
 %endif
 
+# With fedora 33 the POSIX module was split out of the perl
+# package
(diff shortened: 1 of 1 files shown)


dist/build.spec
```diff
@@ -88,6 +88,13 @@ Requires:       %{__pkg_name}-mkbaselibs
 Recommends:     %{__pkg_name}-mkdrpms
 %endif
 
+# With fedora 33 the POSIX module was split out of the perl
+# package
+%if 0%{?fedora_version} >= 33
+BuildRequires: perl(POSIX)
+Requires: perl(POSIX)
+%endif
+
```
darix, line 97: Wouldn't it make sense to always require it? The perl packages should have provides for the intree copies of modules.

PR #584, opened 2020-07-27
support recursive kiwi profile usage

adrianschroeter:
lookup recursive, but avoid endless loop


adrianschroeter:
solved in a different way

PR #585, opened 2020-07-27
build-vm-kvm: on power8 with qemu-5.0 add cap-ccf-assist=off

bugfinder:


bugfinder:
tackling https://bugzilla.suse.com/show_bug.cgi?id=1174374

PR #587, opened 2020-07-30
createzyppdeps repo type is not set anymore, rpm-md is assumed default

adrianschroeter:
boo#1174708


mlandres:
You could as well test for the presence of a `repodata` dir (rpmmd) or a `content` file (yast). Otherwise it's proably a plaindir repo.

adrianschroeter:
we would need to add support for plain dirs first ....

mlandres:
Or let plaindir fail as it does now.

mlschroe:
Fixed in a different way.

PR #592, opened 2020-09-01
Modernize expanddeps

perlpunk:
* Use Getopt::Long (shorter, more readable)
* use defined-or //= (requires perl v5.10)
* Update Copyright


mlschroe:
Do we want to set `require_order` to stay compatible?

perlpunk:
Ok, I added `require_order`

adrianschroeter:
mls thought again and did not like that approach ... I am just the messenger, please shoot him

PR #593, opened 2020-09-01
Add Github Actions

perlpunk:


perlpunk:
for some reason it doesn't trigger a test run. will close and make another PR

perlpunk:
What do I need to do to get this accepted? :)

adrianschroeter:
we run the tests already as part of the package build and I do not really want to get blocked by a ubuntu only test...
IMHO we should finish the obs integration in first place to test all platforms and not only one which is less interessting for us ...

perlpunk:
I don't really understand why automatic CI on a pull request would "block" you.
Can you please share a link where pull requests are currently tested?
How do you test pull requests before they are merged?

Regarding ubuntu vs. other platforms: one can easily switch the container in the github workflow, so I also don't understand why this speaks against having a github workflow.

adrianschroeter:
we by-pass the pull requests here anyway quite often, and we need to test on all distributions and inside of OBS anyway, since the "make test" is testing only a very small subset.

But the next guy (@M0ses) came meanwhile with an example where this could be useful, so let's take it (and avoid the travis dependency at least).

abitrolly:
>  IMHO we should finish the obs integration in first place to test all platforms and not only one which is less interessting for us ...

@adrianschroeter is there an issue open for OBS + GitHub integration?

adrianschroeter:
On Donnerstag, 12. November 2020, 08:01:45 CET Anatoli Babenia wrote:
> >  IMHO we should finish the obs integration in first place to test all platforms and not only one which is less interessting for us ...
> 
> @adrianschroeter is there an issue open for OBS + GitHub integration?

no idea, but an issue is not granting any resources.

It is develpoment work which needs to be done not administrative work ;)

-- 

abitrolly:
@adrianschroeter I expect there are not many ready to use components for such integration. :D From the other side administrative work help with allocating the missing budget. I don't know what is OBS source of income, but a separate private tier on GitHub Marketplace could be the one. )

PR #597, opened 2020-09-08
Support Dockerfile builds with multiple FROM sections

adrianschroeter:
also fixes issue if no RUN line is inside of Dockerfile (section)

Additional perl helper is used now instead of simple sed replacement.

We must enable obs-docker-support after a possible existing "USER root"
line or after each FROM line.

disable must happen before a USER line (except it is the "USER root"
line) or at the end of each section

### DIFF
--- a/Build/Docker.pm
+++ b/Build/Docker.pm
@@ -202,6 +202,7 @@ sub parse {
     @args = split(/[ \t]+/, $line);
     s/%([a-fA-F0-9]{2})/chr(hex($1))/ge for @args;
     if ($cmd eq 'FROM') {
+      shift @args if @args && $args[0] =~ /^--platform=/;
       if (@args && !$basecontainer && $args[0] ne 'scratch') {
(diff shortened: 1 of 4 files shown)


dmarcoux:
With `multiple FROM sections`, are you referring to [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/)?

mlschroe (approved):

adrianschroeter:
@dmarcoux it could be multi-stage builds, but they can also be independend.

eduardoj:
Should have not been the title of the second commit "Support Dockerfile builds with multiple USER sections" instead of "Support Dockerfile builds with multiple FROM sections"?

eduardoj:
> Should have been the title of the second commit "Support Dockerfile builds with multiple USER sections" instead of "Support Dockerfile builds with multiple FROM sections"?

Of course, not. I got it. Each section is created from several `FROM` lines.

PR #601, opened 2020-09-16
build: also export the variables BUILD_USER

bugfinder:
and ABUILD_UID ABUILD_GID


mlschroe:
That won't work as BUILD_USER is not set when init_buildsystem is called.

bugfinder:
mlschroe: better now ?

PR #604, opened 2020-09-17
Add %arm64 from upstream to std_macros

dcermak:
Upstream added `%arm64` in https://github.com/rpm-software-management/rpm/commit/b533164b25052adb78faf6ec57742f280940a172


bugfinder (approved):
thanks

PR #605, opened 2020-09-23
Align test condition with error message in build-vm-ec2

marcus-h:
The error message states that the AWS_ACCESS_KEY and AWS_SECRET_KEY
env variables should be set but the condition only checks for
AWS_ACCESS_KEY (two times). Instead, check for both variables.

Implicit assumption: the error message is correct and the condition
is wrong (I have no clue about AWS - this was just spotted while
reading the code).

### DIFF
--- a/build-vm-ec2
+++ b/build-vm-ec2
@@ -72,7 +72,7 @@ vm_verify_options_ec2() {
     fi
 
     # verify settings
-    if test -z "$AWS_ACCESS_KEY" -o -z "$AWS_ACCESS_KEY" ; then
+    if test -z "$AWS_ACCESS_KEY" -o -z "$AWS_SECRET_KEY" ; then
         cleanup_and_exit 3 "ERROR: No amazon EC2 environment set. Set AWS_ACCESS_KEY and AWS_SECRET_KEY."
     fi
     . /etc/profile.d/ec2.sh


marcus-h:
Hmm the following two checks look a bit strange
```
    if test -z "$BUILD_EC2_AKI" ; then
        cleanup_and_exit 1 "Unknown Amazon EC2 Zone: $BUILD_EC2_ZONE"
    fi
    if test -z "$BUILD_EC2_AKI" ; then
        cleanup_and_exit 3 "ERROR: No image refering to kernel and ramdisk is defined in BUILD_EC2_AKI env."
    fi
```
Probably one of them should be removed?!

adrianschroeter:
On Donnerstag, 24. September 2020, 03:24:32 CEST Marcus Hüwe wrote:
> Hmm the following two checks look a bit strange
> ```
>     if test -z "$BUILD_EC2_AKI" ; then
>         cleanup_and_exit 1 "Unknown Amazon EC2 Zone: $BUILD_EC2_ZONE"
>     fi
>     if test -z "$BUILD_EC2_AKI" ; then
>         cleanup_and_exit 3 "ERROR: No image refering to kernel and ramdisk is defined in BUILD_EC2_AKI env."
>     fi
> ```
> Probably one of them should be removed?!

maybe, but we do not have a test instance atm.

So I'd prefer not to touch it atm until someone has a real setup.
(Or we may drop it entirely, because the last time EC2 just did not
 perform in a reasonable way for us)

-- 

marcus-h:
Fair enough - I'll close this PR for now.

PR #606, opened 2020-09-24
Take .pkgs directory into account in --linksource mode

adrianschroeter:


mlschroe:
This opens up a security hole.

PR #614, opened 2020-11-02
use "zypper download" for zypp:// urls

M0ses:
Without this patch SUSE-connect credentials are not used when downloading
packages from zypp:// urls in a local build.

This patch changes the download mechanism for zypp:// urls from LWP::UserAgent
to system call of "zypper download"

Fixes: #boo1148916, #1174042


M0ses:
Please don`t merge - I will create an OBS server appliance image and test it to make sure that it does not break our appliance

mlschroe (approved):

PR #620, opened 2020-11-10
fix errors while building in obs

M0ses:
OBS needs to know that it must install perl(YAML) otherwise the
test suite fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,4 @@
+language: perl
+perl:
+  - "5.30"
+script: "make test"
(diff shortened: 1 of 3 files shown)


M0ses:
Could we enable travis for this repo to see these type of errors before updating OSU?

adrianschroeter:
the smoke test is executed now via this https://github.com/openSUSE/obs-build/pull/593

M0ses:
closed in favour to github actions #593

PR #625, opened 2020-11-11
fixed BuildRequires for YAML

M0ses:


perlpunk:
Why was `perl(YAML)` added? We're not using it as far as I can see (and we shouldn't).

PR #634, opened 2020-12-04
introduce the option to disable dir_index in ext fs

adrianschroeter:
This will reduce the build speed but make the ordering of entires in
the filesystem reproducible.

It can be enable via CLI switch or via config option. eg

BuildFlags: vmfstype:ext4:nodirindex


adrianschroeter:
@DimStar77 something what you may want to consider for factory once it is merged. (for reproducible builds)

mlschroe (approved):

PR #635, opened 2020-12-07
More README.md edits

abitrolly:
* Add link to latest builds
* Fix typos

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,11 @@ This repository provides the `build` tool to build binary packages in a
 a safe and reproducible way. It can be used standalone or within the
 Open Build Service (OBS).
 
+Latest packages for `obs-build` can be downloaded from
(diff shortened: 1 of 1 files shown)


abitrolly:
It is interesting why LXC is considered insecure?
And if LXD is secure for this purpose?

abitrolly:
How `docker` builds are different that they provide limited reproducibility?

adrianschroeter:
actually, I dunno anymore why LX? had problem some problem, but there is something in my memories. 
It would need a review by security people in any case to declare it secure. But we do not use it atm, so we have no priority on it.

(it is nothing against LXC as technology, but keep in mind that the build happens as root user, so there must be a more complete hardening as when running containers as normal user).

adrianschroeter:
thanks a lot for your fixes!

abitrolly:
Just saving time for future me. )

PR #644, opened 2020-12-15
ccache: it was updated to expect a suffix for -evict-older-than option

sjamgade:
this suffix is mandatory and should either be s(econds)/d(ays).

this patch tries both (with and without suffix), avoiding to do any
complex checks for version (ccache -Version). The version comparison
would need either need floating point comparision or extracting major
and minor version and comparing those.

The with_suffix variant is tried first as it is what will be supported
going ahead.

Would be nice to have a second pair of eyes for the bash !

### DIFF
--- a/build
+++ b/build
@@ -1611,8 +1611,11 @@ if test -n "$CCACHE" ; then
     if test -n "$REMOVE_CCACHE" ; then
         echo "... cleaning ccache"
         test_cmd="ccache -h | grep -c evict-older-than"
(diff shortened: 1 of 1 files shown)


mlschroe:
Hi Sumit!

Your old patch used atoi() on the argument, right? So couldn't we always add the "s" suffix (which atoi() will ignore)?

sjamgade:
Hi Michael, 
just checked `atoi`'s man page. yes we can !

PR #646, opened 2020-12-19
debtransform: Add Debian revision if not present

### LINKED ISSUE #633: debtransform: produces broken builds for Ubuntu 20.10
`debtransform` converts native packages to **non-native** by creating `.orig.tar.gz` file.

```
[  101s] Found files matching debian.*, running debian transformer...
[  101s] release: (4.1), release (DEB) (4.1)
[  101s] ** Started: debtransform --release 4.1 //usr/src/packages/SOURCES //usr/src/packages/SOURCES/bleachbit.dsc //usr/src/packages/SOURCES.DEB
[  101s] No DEBTRANSFORM-TAR line in the .dsc file.
[  101s] Attempting automatic discovery of a suitable source archive.
[  101s] Source archive chosen for transformation: bleachbit-4.1.1.tar.gz
[  101s] No DEBTRANSFORM-FILES-TAR line in the .dsc file.
[  101s] Attempting automatic discovery of a debian archive.
[  101s] Hardlinking //usr/src/packages/SOURCES/bleachbit-4.1.1.tar.gz to //usr/src/packages/SOURCES.DEB/bleachbit_4.1.1.orig.tar.gz
[  101s] files d7dd9b320579971308c78245c0a3ff2f 1147196 bleachbit_4.1.1.orig.tar.gz
[  101s] Scanning //usr/src/packages/SOURCES.DEB/bleachbit_4.1.1.orig.tar.gz...
[  101s] Generating //usr/src/packages/SOURCES.DEB/bleachbit_4.1.1.diff
[  101s] Processing file "//usr/src/packages/SOURCES/debian.rules"...
[  101s] Processing file "//usr/src/packages/SOURCES/debian.control"...
[  101s] Processing file "//usr/src/packages/SOURCES/debian.changelog"...
[  101s] Writing //usr/src/packages/SOURCES.DEB/bleachbit_4.1.1.dsc
```

This makes package version invalid, because non-native packages should have a suffix. Newer versions of `dpkg-source` make this an error.
```
[  103s] dpkg-source: error: non-native package version does not contain a revision
```

`debtransform` should either append suffix automatically, or don't convert the package from native.

abitrolly:
Fixes #633. Debian revision is required for non-native packages, but the force check was added only in Ubuntu 20.10.


nickbroon:
Thanks for merging this @adrianschroeter :-)
Is a release on obs-build and then it's deployment to the OBS servers planned?

sni:
This seem to have broken debian testing builds where debian revisions are not allowed for native packages.

```
[   69s] ** Started: debtransform --release 23.35 //usr/src/packages/SOURCES //usr/src/packages/SOURCES/naemon.dsc //usr/src/packages/SOURCES.DEB
[   69s] No DEBTRANSFORM-TAR line in the .dsc file.
[   69s] Attempting automatic discovery of a suitable source archive.
[   69s] Source archive chosen for transformation: naemon-1.2.2.0+20201113~fd83da1.tar.gz
[   69s] No DEBTRANSFORM-FILES-TAR line in the .dsc file.
[   69s] Attempting automatic discovery of a debian archive.
[   69s] Transforming into source package 1.0 (non-native) format
[   69s] Added Debian revision to Version field, which is now "1.2.2.0+20201113~fd83da1-1".
[   69s] Modifying dsc Version field to "1.2.2.0+20201113~fd83da1-1"
...
[   70s] dpkg-source: error: can't build with source format '3.0 (native)': native package version may not have a revision
[   70s] dpkg-buildpackage: error: dpkg-source -b . subprocess returned exit status 25
```

I'd suggest `debtransform` to set the deb source format to '3.0 (native)' and remove adding a revision.

abitrolly:
@sni `debtransform` only generates [1.0 non-native source](https://manpages.debian.org/buster/dpkg-dev/dpkg-source.1.en.html) format, with `.diff.gz` file etc., where debian revision is required by the spec. Do you have the `Format` field set to `3.0 (native)` in your `.dsc` files? In that case I am afraid the fix for `debtransform` would be to reset it to `1.0` explicitly.

abitrolly:
Actually, `debtransform` already resets the format field to 1.0

https://github.com/abitrolly/obs-build/blob/1d5952c692c233b18109ed557f82837815508012/debtransform#L416

Something else is going on here.

sni:
My source debian folder from the tarball contains a `debian/source/format` file containing `3.0 (native)`. That seems to override the format version from the dsc file. Is there any way to remove/override or overwrite that file from the `_service`?

abitrolly:
@sni it is possible. Original tarball is repacked anyway if it is not` .tar.gz`. Why you using the `debstransform` in the first place? For native packages with `debian/` directory sources OBS should build it them. Just remove **debian.*** files so that `debtransform` doesn't trigger.

sni:
@abitrolly i assume debtransform is automatically triggered from setversion. For example here:
https://build.opensuse.org/package/live_build_log/home:naemon:daily/naemon/Debian_Testing/x86_64

abitrolly:
@sni it looks like https://build.opensuse.org/package/view_file/home:naemon:daily/naemon/_service is extracting the `debian/*` into `debian.*` files. Try to remove those blocks and then `debtransform` won't trigger and OBS will reuse `debian/` info from the archive.

sni:
If i remove those files from the _service it won't build at all with the message:
```
The package build has been disabled in package build description (for example in the .spec file) or does not provide a matching build description for the target.
```
Btw, i do not extract all `debian/` files, only a few of them which seemed to be necessary (and debian/source/format is not one of them).
But yes, that was the idea back then, have a `debian/` folder in the source tarball and keep the `_service` as simple as possible. Basically setting a new version rebuild the tarball.

abitrolly:
@sni you need to make sure that `.dsc` file is present. This should not be removed.

```
  <service name="extract_file">
    <param name="archive">*.tar</param>
    <param name="files">*/debian/naemon.dsc</param>
    <param name="outfilename">naemon.dsc</param>
  </service>
```

sni:
nope, that's not working. I need debtransform to change the version and recreate the tarball. Not running debtransform is not an option. At least i don't see how this could work.
```
[   81s] dpkg-source: error: cannot fstat file /usr/src/packages/SOURCES/naemon_1.0.0.orig.tar.gz: No such file or directory
```

Basically it worked great before this change. But i am happy with whatever solution that works. Do you have any example by chance for the following workflow:
1. clone git repository (containing source and debian folder and spec file)
2. set version based on pattern
3. build packages / repositories

It worked exactly like expected before this change. I don't want to manually maintain the builds every day.

abitrolly:
Why do you need https://github.com/naemon/naemon/blob/master/debian/source/format then?

sni:
indeed, i don't think it is required for obs builds. Is there a way to remove that file during debtransform or rebuilding the tarball or fetching the sources or anywhere in the _services run? Then i can dry a  testrun.

Besides that, could the appending `-1` to the version be made optional? Or it could be added to the `setversion` like 
```
<param name="versionformat">@PARENT_TAG@.@TAG_OFFSET@+%cd~%h-1</param>
```
if someone wanted this.

abitrolly:
@sni the official docs use `obs_scm` + `set_version` for cloning. I am not sure how they are supposed to interact with `debtransform`. Must admit I am not an expert in OBS and Debian packaging at all. Need more input from OBS maintainers, and it is better to do this in a separate issue.

There is no clear docs how to use `obs_scm` with Debian packaging.

https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.best-practices.scm_integration.html

https://forums.opensuse.org/showthread.php/548839-Newbie-question-get-debian-directory-from-git-repo

abitrolly:
> Besides that, could the appending -1 to the version be made optional? 

No. That's the reason this PR was created in the first place. The check that non-native Debian packages should contain Debian revision is enforced in newer `dpkg-source` and that broke Ubuntu 20.10 builds for many packages, including openSUSE tools themselves. `debtransform` only generates non-native source package. For beautiful version, the `debian/` dir in your repo should have the native source layout, and then there should be some doc how to update version for new releases. Following the doc it might be possible to define OBS service that does what you want.

PR #647, opened 2021-01-05
Improve 'build' and 'osc build' experience

e4t:
This set contains two things:

1. Improvement of the 'build' and 'osc build' experience: together with changes to github.com/openSUSE/osc and if supported for the underlying build system, allow for 'partial' builds - like rpmbuild --short-circuit -b<build_stage>, but better: for multi-stage builds, it allows to execute:
    - a single build stage, 
    - a build up to a specified stage or 
    - a build starting from a specified stage (provided all previous stages have already been performed).
No more waiting for endless %builds just to test a change to an %files stage, no more need to fiddle with rpmbuild directly in the chroot (and forgetting to copy back the changes to the spec file later).
(Note: this is presently available for rpmbuild only, the design is sufficiently generic that it should be implementable to every other build system - provided this supports multiple build stages and running them individually.
2. Additions to the documentation to document the changes in (1.) as well as some other missing bits and pieces.

### DIFF
--- a/build-recipe-spec
+++ b/build-recipe-spec
@@ -21,13 +21,75 @@
 #
 ################################################################
 
+# Check build stage and set stages list according to BUILD_RPM_BUILD_MODE:
+# there are three modes:
+# 'pre' or not set: build all stages up to BUILD_RPM_BUILD_STAGE
+#                  if not set, defaults to -ba
+# 'single': do an rpmbuild --short-circuit' build on stage BUILD_RPM_BUILD_STAGE
+# 'post': continue building from stage BUILD_RPM_BUILD_STAGE
+
+rpm_build_keep=
+rpm_build_stages=
+
+check_set_build_stages() {
+    local pre_d post_s
+    test -n "$BUILD_RPM_BUILD_MODE" || BUILD_RPM_BUILD_MODE=pre
+    test -z "$BUILD_RPM_BUILD_STAGE" -a "$BUILD_RPM_BUILD_MODE" != "pre" && \
+       cleanup_and_exit 1 "Error: --buildmode only allowed with --stage or BUILD_RPM_BUILD_STAGE set"
+    test -z "$BUILD_RPM_BUILD_STAGE" && BUILD_RPM_BUILD_STAGE=-ba
+    case $BUILD_RPM_BUILD_MODE in
+       pre) ;;
+       single|post)
+           test "$BUILDTYPE" = debbuild && cleanup_and_exit 1 "Error: --mode not implemented for debbuild"
+           test -z "$BUILD_RPM_BUILD_STAGE" && \
+               cleanup_and_exit 1 "Error: --buildmode only allowed with no --stage setting";
+           test "$CLEAN_BUILD" != "--clean" || \
+                             cleanup_and_exit 1 "Error: --clean not compatible with --mode"
+           case $BUILD_RPM_BUILD_STAGE in
+               -bp) test $BUILD_RPM_BUILD_MODE = post && BUILD_RPM_BUILD_STAGE=-ba
+                    BUILD_RPM_BUILD_MODE=pre ;;
+               -bc) pre_d="BUILD"; post_s="-bc -bi -bb -bs" ;;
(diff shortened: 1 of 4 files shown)


mlschroe:
Hmm, can't say I like the user interface with the new 'mode' option. And I also dislike the old '--stage' option, as you always need to prefix the stage with `-b`.
I would rather have extend --stage so that you it offers to specify a range, like:
```
--stage=i    only do the install stage
--stage=-i   do all stages up to install
--stage=i-   do stages starting from install
```
We can also use some different char for the range, so that it does not look that much like an option.

Of course we need to stay backward compatible, so that --stage=-bX still works.

build-recipe-spec
```diff
@@ -21,13 +21,75 @@
...
+           test "$BUILDTYPE" = debbuild && cleanup_and_exit 1 "Error: --mode not implemented for debbuild"
+           test -z "$BUILD_RPM_BUILD_STAGE" && \
+               cleanup_and_exit 1 "Error: --buildmode only allowed with no --stage setting";
+           test "$CLEAN_BUILD" != "--clean" || \
+                             cleanup_and_exit 1 "Error: --clean not compatible with --mode"
+           case $BUILD_RPM_BUILD_STAGE in
+               -bp) test $BUILD_RPM_BUILD_MODE = post && BUILD_RPM_BUILD_STAGE=-ba
+                    BUILD_RPM_BUILD_MODE=pre ;;
+               -bc) pre_d="BUILD"; post_s="-bc -bi -bb -bs" ;;
+               -bi) pre_d="BUILD"; post_s="-bi -bs -bb" ;;
```
mlschroe, line 52: -bs should always come atfer -bb

build-recipe-spec
```diff
@@ -126,91 +187,97 @@ recipe_build_spec() {
...
-    toshellscript $rpmbuild \
-	    "${definesnstuff[@]}" \
-	    "${rpmbopts[@]}" \
-	    "$TOPDIR/SOURCES/$RECIPEFILE" \
-	    > $BUILD_ROOT/.build.command
-    chmod 755 $BUILD_ROOT/.build.command
-    check_exit
-    chroot $BUILD_ROOT su -c /.build.command - $BUILD_USER < /dev/null
-    st=$?
+	test "$st" = 0 -o "$st" = 11 || break
```
mlschroe, line 279: Why do you also test for `11` here?

e4t: Right, on 11 the caller should probably handle this. I will fix this.

build-recipe-spec
e4t: Ok. will fix.

e4t:
I'll change the semantics of --stage according to Michael's suggestion (of course I will keep the old semantics intact).

e4t:
I've updated the patch. Instead of a `--mode=...` option, I've introduced 3 new forms of the `--stage` option:
- `--stage=..STAGE` to build up to STAGE,
- `--stage=STAGE` to build only STAGE,
- `--stage=STAGE..` to build everything starting from STAGE.

All other issues were also addressed.

mlschroe:
OK, I somewhat rewrote the code and manually merged it. I did change the stage selection, because I think that specifying a single letter stage should to the expected thing, i.e. just pass -bSTAGE to rpm and build the stages up to and including the specified stage.

So I needed to come up with ways to select "just this stage" and "this stage and the stages after it". I went with a trailing '+' to specify the latter and a trailing '=' to specify single stage builds. I'm not super happy about the '=', though. And I'm also wondering if we should have a 'everything after this stage excluding the specified stage' flag, because i can imaging that this is more useful than what we have. Suggestions welcome.

Anyway, thank you very much for opening this pull request!

e4t:
I personally have the greatest need for this for the the %install stage. The install stage sets the 'stage' for everything that comes later. If a file is moved during %install my %files sections may no longer work. Technically, these things usually get fixed by adjustments to the %files section, however, running the this section without doing a prior install usually fails. However, %install takes only a fraction of the time of %build usually, so this is not doing exactly what I need. I'm fine with it.

Thanks for looking into it!

PR #649, opened 2021-01-16
Support autosetup and autopatch

bmwiedemann:
Without this patch,
%autosetup caused it to die in line 131
and %autopatch succeeded but would fail to build
because it applied the patch twice.

Might fix https://bugzilla.opensuse.org/show_bug.cgi?id=1172891

I tested this patch on `%setup`, `%autosetup` and `%autopatch` cases.

The -p level is ignored


adrianschroeter:
I wonder why this script is part of the build script code at all... we seem not to use it.

the mentioned use case  in the bugreport is independend of this script at least AFAIK.

PR #654, opened 2021-01-22
spec: Remove YAML.pm requirement

perlpunk:
We're not using it anywhere, and we shouldn't, since it's outdated and
broken.


adrianschroeter:
verifing in home:adrianSuSE:branches:openSUSE:Tools/build  atm

PR #656, opened 2021-01-27
Support execution of rpmlint-image.

marxin:
The following pull request will enable `rpmlint-image` package to be used for the upcoming rpmlint2.

The patch ignores `/opt/testing/bin/rpmlint` and detects if `rpmlint-image` is present. If so, then `*.rpm` files (and `*rpmlintrc`) are copyed into the `rpmlint-image` chroot.

I've tested that locally in `https://build.opensuse.org/project/show/home:marxin:rpmlint-playground`.

### DIFF
--- a/build-recipe-spec
+++ b/build-recipe-spec
@@ -252,31 +252,48 @@ recipe_check_file_owners() {
 }
 
 recipe_run_rpmlint() {
-    # SUSE builds are using a special rpmlint binary which is not
-    # pulling in additional dependencies
-    local rpmlint="/opt/testing/bin/rpmlint"
-    if ! test -x "$BUILD_ROOT$rpmlint" ; then
-        # default rpmlint place as fallback
-        rpmlint="/usr/bin/rpmlint"
-    fi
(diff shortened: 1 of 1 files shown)


adrianschroeter:
Please put the code copying in some package. Code in build script needs to be fully forward and backward compatible for 20 years. IMHO it is unlikely that this code must not be touched again.

marxin:
> But why can't you just use the post-build checks hooks which do already exist?

I can add `rpmlint2` execution as one another post-build check.
How will we disable rpmlint to be run in this script?

adrianschroeter:
On Mittwoch, 27. Januar 2021, 10:55:03 CET marxin wrote:
> > But why can't you just use the post-build checks hooks which do already exist?
> 
> I can add `rpmlint2` execution as one another post-build check.
> How will we disable rpmlint to be run in this script?

well, we would just drop the installation of rpmlint in the base project build configuration
and it won't be there.

No?

-- 

marxin:
> well, we would just drop the installation of rpmlint in the base project build configuration and it won't be there. No?

Yes, it should help! Thank you for the hint.

PR #664, opened 2021-02-16
support local file repositories for pbuild

adrianschroeter:


adrianschroeter:
retracting, as it makes the code slightly less secure.

PR #665, opened 2021-02-23
Set LC_ALL to C.UTF-8

rickysarraf:
Multiple packages have run into build failures on our OBS. On closer
inspection, it was concluded to be issues with the locale settings.
These issues were never seen on the Debian buildd infrastructure because
there the locale setting `LC_ALL=C.UTF-8` is in use.

This change keeps us closer to what is used in Debian buildds.

Signed-off-by: Ritesh Raj Sarraf <ritesh.sarraf@collabora.com>

### DIFF
--- a/build-recipe-dsc
+++ b/build-recipe-dsc
@@ -95,7 +95,7 @@ dsc_build() {
     # this allows the build environment to be manipulated
     # and alternate build commands can be used
     DSC_BUILD_CMD="$(queryconfig --dist "$BUILD_DIST" --archpath "$BUILD_ARCH" --configdir "$CONFIG_DIR" substitute dsc:build_cmd)"
-    test -z "$DSC_BUILD_CMD" && DSC_BUILD_CMD="dpkg-buildpackage -us -uc"
+    test -z "$DSC_BUILD_CMD" && DSC_BUILD_CMD="LC_ALL=C.UTF-8 dpkg-buildpackage -us -uc"
     if test -e $buildroot/$TOPDIR/SOURCES/build.script ; then
 	echo "Sourcing build.script to build - it should normally run 'dpkg-buildpackage -us -uc'"
 	DSC_BUILD_CMD="source $TOPDIR/SOURCES/build.script"


rickysarraf:
Multiple build failures, relevant to this change.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=982974
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=981718
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=981847

mlschroe:
C.UTF-8 does not exist in older distros, so I'm not sure if it is a good idea to set it unconditionally.

coolo:
https://unix.stackexchange.com/questions/597962/how-widespread-is-the-c-utf-8-locale suggests Arch Linux does not have it - so it's not about the age of the distro

mlschroe:
True, but this pull request just patches the build-recipe-dsc file...

coolo:
but it's about the distribution executing it, or do I get this wrong?

andrewshadura:
In this case the change only affects `DSC_BUILD_CMD` which run in a Debian (or a Debian derivative) chroot, so what other distributions do isn’t relevant.

build-recipe-dsc
```diff
@@ -95,7 +95,7 @@ dsc_build() {
     # this allows the build environment to be manipulated
     # and alternate build commands can be used
     DSC_BUILD_CMD="$(queryconfig --dist "$BUILD_DIST" --archpath "$BUILD_ARCH" --configdir "$CONFIG_DIR" substitute dsc:build_cmd)"
-    test -z "$DSC_BUILD_CMD" && DSC_BUILD_CMD="dpkg-buildpackage -us -uc"
+    test -z "$DSC_BUILD_CMD" && DSC_BUILD_CMD="LC_ALL=C.UTF-8 dpkg-buildpackage -us -uc"
```
em-, line 98: Is this the right place?

Shouldn't we change [`build-pkg-deb`](https://github.com/openSUSE/obs-build/blob/9f60ac1fa34fca090dd16dda9c21ba16769830a1/build-pkg-deb#L40) instead?

(but I am not sure who calls who to be honest)

andrewshadura: Probably both need to be changed; also `build-pkg-deb` needs to consistently run `deb_chroot`.

andrewshadura:
Any update on this?

rickysarraf:
Also added a couple of fixes we've been carrying in our setup for a while.

obbardc:
@rickysarraf do you have any time to resolve this ? It would be nice to have this in OBS by default.

nickbroon:
For additional context on Debian buildd setting LC_ALL to C.UTF-8 see: https://lists.debian.org/debian-devel/2024/06/msg00188.html

The lack of OBS also doing this also appears to be cause or recent breakage of installing fontconfig in OBS build environments reported here: https://lists.opensuse.org/archives/list/buildservice@lists.opensuse.org/thread/C24P565DGJLSB4VDKHZVC7YK5KMQYUCP/

As such I would support the merging of pull request.

andrewshadura:
@rickysarraf, there are a few more related changes in https://gitlab.collabora.com/obs/obs-build/-/commits/collabora/main, could you please pick them up too?

rickysarraf:
> @rickysarraf, thanks! There are a few more related changes in https://gitlab.collabora.com/obs/obs-build/-/commits/collabora/main, could you please pick them up too?

@andrewshadura Sure. I'm on it. Will check now and pick the ones relevant.

andrewshadura:
I think it’s better to leave out 4e273c6. This was fixed in a different way upstream.

obbardc (approved):
Looks good to me.

I will test this in more detail with the failing Debian packages later, to see if it fixes the problem.

build-recipe-dsc
obbardc: Think this is OK now, right ?

andrewshadura:
@mlschroe ↑

PR #670, opened 2021-03-01
source common_functions early

marcus-h:
vm_detect_2nd_stage can call cleanup_and_exit before setting
RUNNING_IN_VM=true. Thus, buildroot_umount, which is defined in the
common_functions file, can be called by cleanup_and_exit. Hence,
source the common_functions file before calling vm_detect_2nd_stage.

Note: if sourcing the common_functions file fails, we still call
cleanup_and_exit and the umount fails. However, all other cleanup
actions are executed.

### DIFF
--- a/build
+++ b/build
@@ -838,6 +838,8 @@ umount_stuff() {
 
 trap fail_exit EXIT
 
+. $BUILD_DIR/common_functions || exit 1
+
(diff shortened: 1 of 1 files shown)


darix:
verified

PR #671, opened 2021-03-01
add --no-clean option to re-use buildroot

adrianschroeter:


PBuild/Options.pm
```diff
@@ -199,6 +201,8 @@ Important options (see the man page for a full list):
         defaults to the host architecture
   --obs url
         open build service instance for obs:/ type urls
+  --no-clean
+        re-use former build environment
```
mlschroe, line 205: Please do not mention no-clean in the usage. The man page is enough.

adrianschroeter: done

PR #675, opened 2021-03-11
add generic post build check script hook

adrianschroeter:
Execute every script inside of /usr/lib/build/post-build-checks/ .

The check must validate that it runs for the right build type and fail
only when it finds a problem with a build result. No matter of which
type.

We already have usr/lib/build/checks/ but the existing checks are rpm
generic and fail on non rpm builds. Ideally these get migrated later on.

OBS-110

### DIFF
--- a/build
+++ b/build
@@ -1649,6 +1649,13 @@ if test -n "$RPMS" -a -d "$BUILD_ROOT/usr/lib/build/checks" ; then
     test -e "$BUILD_ROOT/proc/self" || mount -n -tproc none $BUILD_ROOT/proc
 fi
 
+for CHECKSCRIPT in $BUILD_ROOT/usr/lib/build/post-build-checks/* ; do
+    if test -x "$CHECKSCRIPT"; then
+       echo "... running ${CHECKSCRIPT##*/}"
+       BUILD_ROOT=/ chroot "$BUILD_ROOT" "/usr/lib/build/post-build-checks/${CHECKSCRIPT##*/}" || cleanup_and_exit 1
+    fi
+done
+
 # checkscripts may have deleted some binaries
 RPMS=`find $BUILD_ROOT/$TOPDIR/RPMS -type f -name "*.rpm" 2>/dev/null || true`
 DEBS=`find $BUILD_ROOT/$TOPDIR/DEBS -type f -name "*.deb" 2>/dev/null || true`


Vogtinator:
Can you provide an example of how such checks would look like, especially for docker/kiwi builds? For container builds I'm especially wondering how to access the build image with e.g. podman/skopeo.

> The check must validate that it runs for the right build type and fail only when it finds a problem with a build result. No matter of which type.

How do they know what the current build type is?

https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.qa_hooks.html#id-1.5.10.17.6.4 doesn't really have much info...

adrianschroeter:
Check the post-build-checks package for the rpm checks. You can find the local generated files below $BUILD_ROOT$TOPDIR/DOCKER/

TOPDIR can be set like this:

TOPDIR=/usr/src/packages
test -d $BUILD_ROOT/.build.packages && TOPDIR=/.build.packages
test -L $BUILD_ROOT/.build.packages && TOPDIR=/$(readlink $BUILD_ROOT/.build.packages)

also the local registry should still run (complain if not). But it does not contain the build result. However, the build script could upload it to the local registry if needed.

Not sure what exactly you want to test? For used packages/licenses the .packages file might be more useful though.

Vogtinator:
> Check the post-build-checks package for the rpm checks. You can find the local generated files below $BUILD_ROOT$TOPDIR/DOCKER/

Also for kiwi builds producing container images?

> Not sure what exactly you want to test? For used packages/licenses the .packages file might be more useful though.

What is currently needed the most is a validator for metadata, i.e. image names/tags and labels. I'm sure that in the future we'll probably want to look inside the image as well, maybe to check for anything with odd ownership/permissions or similar issues.

adrianschroeter:
On Donnerstag, 11. März 2021, 16:30:45 CET Fabian Vogt wrote:
> > Check the post-build-checks package for the rpm checks. You can find the local generated files below $BUILD_ROOT$TOPDIR/DOCKER/
> 
> Also for kiwi builds producing container images?

no, it is in KIWI directory there indeed

> > Not sure what exactly you want to test? For used packages/licenses the .packages file might be more useful though.
> 
> What is currently needed the most is a validator for metadata, i.e. image names/tags and labels. I'm sure that in the future we'll probably want to look inside the image as well, maybe to check for anything with odd ownership/permissions or similar issues.

For that usecase the .docker.containerinfo file might be helpful?


-- 

Adrian Schroeter ***@***.***>

Vogtinator:
> On Donnerstag, 11. März 2021, 16:30:45 CET Fabian Vogt wrote: > Check the post-build-checks package for the rpm checks. You can find the local generated files below $BUILD_ROOT$TOPDIR/DOCKER/ Also for kiwi builds producing container images?
> no, it is in KIWI directory there indeed

Some example on how this can be used would be very helpful, also how to get the output into where OBS displays the rpmlint result.

> > Not sure what exactly you want to test? For used packages/licenses the .packages file might be more useful though. What is currently needed the most is a validator for metadata, i.e. image names/tags and labels. I'm sure that in the future we'll probably want to look inside the image as well, maybe to check for anything with odd ownership/permissions or similar issues.
> For that usecase the .docker.containerinfo file might be helpful?

Indirectly yes, by getting the `imageid` from there and then dissecting the matching json blob manually. It would be much easier (and for other tests even required) to to access the built container(s) with docker/podman or umoci for inspection. Otherwise it'll be really painful to deal with layers and all  that.

PR #678, opened 2021-03-18
Build as user support

adrianschroeter:
Only when using ext fs and for KVM.


marcus-h:
On 2021-03-18 17:36:05 +0000, Adrian Schröter wrote:
> diff --git a/build-vm b/build-vm
> index 034e43a8..56190db3 100644
> --- a/build-vm
> +++ b/build-vm

...

> @@ -739,20 +741,19 @@ vm_setup() {
>  	    rm -f "$VM_ROOT"
>  	fi
>      fi
> -    # using virt-make-fs for ext3 file VMDISK_FILESYSTEM
> -    # so nothing is needed to do part from creating appropriate folders
> -    # and calling virt-make-fs accordingly
> -    if test -n "$VM_USE_VIRT_FS" ; then
> -        echo "Using virt-make-fs to create root image"
> +    # Check for populating filesystem during mkfs
> +    if test -z -o "${VMDISK_FILESYSTEM:0:3}" = "ext" ; then

test -z "$VM_POPULATE_FILESYSTEM" -o ... ?

> +        echo "Populating filesystem during formating"

marcus-h:
On 2021-03-18 12:54:08 -0700, Marcus Hüwe wrote:
> On 2021-03-18 17:36:05 +0000, Adrian Schröter wrote:
> > diff --git a/build-vm b/build-vm
> > index 034e43a8..56190db3 100644
> > --- a/build-vm
> > +++ b/build-vm
>
> ...
>
> > @@ -739,20 +741,19 @@ vm_setup() {
> >  	    rm -f "$VM_ROOT"
> >  	fi
> >      fi
> > -    # using virt-make-fs for ext3 file VMDISK_FILESYSTEM
> > -    # so nothing is needed to do part from creating appropriate folders
> > -    # and calling virt-make-fs accordingly
> > -    if test -n "$VM_USE_VIRT_FS" ; then
> > -        echo "Using virt-make-fs to create root image"
> > +    # Check for populating filesystem during mkfs
> > +    if test -z -o "${VMDISK_FILESYSTEM:0:3}" = "ext" ; then
>
> test -z "$VM_POPULATE_FILESYSTEM" -o ... ?
>
Hmm or rather without the "-z ..." check?

adrianschroeter:
@bmwiedemann you worked into that direction before, do you think that removing the default to run_as=qemu has any downside?

My goal here is not to need to ask anymore any root passwort and the switch to qemu user is breaking this.

bmwiedemann:
If the whole script is already running as non-root, the `run_as=qemu` is not as needed.
In theory if something breaks out of the VM, it could get access to more valuable user-data, but IMHO kvm-breakout bugs are rare enough these days.

build-vm-kvm
```diff
@@ -275,9 +276,6 @@ vm_startup_kvm() {
     qemu_args=(-drive file="$VM_ROOT",format=raw,if=none,id=disk,cache=unsafe -device "$kvm_device",drive=disk,serial=0)
     if [ -n "$VM_USER" ] ; then
 	getent passwd "$VM_USER" > /dev/null || cleanup_and_exit 3 "cannot find KVM user '$VM_USER'"
-    else
-	# use qemu user by default if available
-	getent passwd qemu >/dev/null && VM_USER=qemu
```
bmwiedemann, line 280: we could instead wrap this in a `if [ $UID = 0 ] ; then`

adrianschroeter: yep, we will add that

PR #687, opened 2021-05-10
enforce use of runc

dirkmueller:
on a tumbleweed host, podman defaults to use crun, which is able to deal with the enabled unified cgroupv2. However older versions of podman (like SLE) don't have support. so detect in which situation we are, and hardcode runc if crun is not available inside the buildroot.

### DIFF
--- a/call-podman
+++ b/call-podman
@@ -77,7 +77,9 @@ fi
 # setup mounts
 test -e "$BUILD_ROOT/proc/self" || mount -n -tproc none $BUILD_ROOT/proc
 
+# use runc if crun is not available (SLE..)
+RUNTIME_OPT=""
+[ -x "$BUILD_ROOT/usr/bin/crun" ] || RUNTIME_OPT="--runtime runc"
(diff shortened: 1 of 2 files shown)


call-podman
```diff
@@ -77,7 +77,9 @@ fi
 # setup mounts
 test -e "$BUILD_ROOT/proc/self" || mount -n -tproc none $BUILD_ROOT/proc
 
-# run the command
-exec chroot "$BUILD_ROOT" podman --cgroup-manager=cgroupfs "$@"
-
+# use runc if crun is not available (SLE..)
```
Vogtinator, line 80:
```suggestion
# If the host kernel defaults to cgroupsv2, podman tries to run `crun` instead, even if not available.
# As a workaround, force use of runc.
```

Vogtinator (approved):
It's a mystery why my PR fixed podman builds. I know that they failed with systemd errors before it got merged and worked afterwards...

PR #689, opened 2021-05-10
build-recipe-dsc: support optional "OBS-DCH-RELEASE: 1" also in …

jimklimov:
…debtransform-ed sources to append an OBS release number during respins

This allows automatically rebuilt packages (same source and recipe, rebuild happens because of manual trigger or changed dependencies and config requesting to rebuild due to those) to be discernible from earlier built artifacts by virtue of changed version. This is optional, to be used at OBS packagers' discretion, because it goes against reproducible builds (same source and same dependencies and frozen timestamps might produce identical binaries otherwise), but for others it allows to see that a package was "respun" due to changed circumstances. 

It may make sense to augment (or replace?) this option with some pkgconf setting, e.g. to enable such behavior in development project and disable it in release using the same (copied or obs-branched) sources and recipes. For the starting point, I used the same syntax (magic line in recipe file) as already processed for recipes without explicit debian content.

Tested that this also changes the version reflected in the debian.changelog, although debtransform only prepends one entry about itself to the content that may be available in the uploaded recipe and so the *whole* history of respin numbers for unmodified recipe is not known - just the latest build:

````
packagename (0.1.2-1+5.6) stable; urgency=low

  * version number update by debtransform

 -- debtransform <build@opensuse.org>  Mon, 10 May 2021 20:12:02 +0000


packagename (0.1.1-1) stable; urgency=low

  * Initial commit
...
````

Thanks to @adrianschroeter for pointing from IRC in the right direction :)

### DIFF
--- a/build-recipe-dsc
+++ b/build-recipe-dsc
@@ -59,6 +59,25 @@ recipe_prepare_dsc() {
         if [ "$RELEASE" ]; then
           echo "release: ($RELEASE), release (DEB) ($DEB_RELEASE)"
                  RELEASEARGS="--release $DEB_RELEASE"
+          if grep -Eq '^OBS-DCH-RELEASE: 1' $BUILD_ROOT$TOPDIR/SOURCES/$RECIPEFILE \
+          && grep -Eq '^Version:' $BUILD_ROOT$TOPDIR/SOURCES/$RECIPEFILE \
+          ; then
+            WSCHAR="`printf '\t '`"
+            # https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-version
+            VERSION="`grep -E '^Version:' $BUILD_ROOT$TOPDIR/SOURCES/$RECIPEFILE | sed -e 's,^Version:'"[$WSCHAR][$WSCHAR]"'*,,' -e 's,'"[$WSCHAR]"'*$,,'`"
+            echo "OBS-DCH-RELEASE: DSC file originally spelled package version as ${VERSION}"
+            case "$VERSION" in
+                *-*) # Got a debver or more already, add a plus-suffix to debver
+                    OBS_DCH_RELEASE="+$DEB_RELEASE" ;;
+                *)   # No debver, add the first such token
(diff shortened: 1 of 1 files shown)


darix:
Just curious ... we already had DEBTRANSFORM-RELEASE as tag for this. wouldnt this have been enough? 

Right now we have the same code twice for this. Though debransform lacks the whole +/- matching code which could be ported over.

https://github.com/openSUSE/obs-build/blob/master/debtransform#L387

darix:
And we should patch the distribution tag to make lintian happy

jimklimov:
Good point actually. I went over recipes I have nearby, for the on-premise OBS this was developed for initially, and while they did have a fair share of DEBTRANSFORM-TAR lines (usually generated into their DSC files by our package publishing scripts and/or coming from https://github.com/zeromq/zproject generated recipes in the original sources), but not a DEBTRANSFORM-RELEASE as far as I can see.

The DSC files in OBS workspaces also have the `OBS-DCH-RELEASE: 1` lines in them (probably added by the publishing scripts), so I went for that as the trigger for the feature in this PR. And also probably because I missed the other setting.

If you are sure that augmenting `DEBTRANSFORM-RELEASE` is the more-correct way to go, feel free to fix that; just please let us leaf consumers know (link here? in project Changelog?) if recipes would no longer have the automatic `+respin` naming effect due to presence of `OBS-DCH-RELEASE: 1` and would need to add `DEBTRANSFORM-RELEASE` lines (or is there more syntax to it? a `: 1` trailing?)

darix:
@mlschroe  and me looked into this because I showed him my new blog post https://nordisch.org/posts/the-debian-changelog-and-obs/ .

We discussed how to implement patching the debian/changelog file so it contains the correct distribution name, to make lintian happy.

while reviewing what kind of code we already had for this feature, we noticed that this feature is duplicated. So I pinged you to get your feedback.

darix:
I have opened https://github.com/openSUSE/obs-build/pull/818 to unify the patching.

PR #690, opened 2021-05-11
Support loading multiple containers for multi-stage builds (OBS-68)

dirkmueller:
* Consistently use call-podman also for image import
* Code cleanup around docker/podman
* Load all container images with manifests


adrianschroeter:
it seems to be a bit to invasive, can we speak about it in our channel please?

mlschroe:
No way will this go in with one big commit.

Issue #695, opened 2021-06-01
prjconf `Release:` breaks debs’ debian revision


drmuey:
**TL;DR**:

Doing the classic `Release: %%{?release_prefix}.<CI_CNT>.<B_CNT>` in prjconf breaks debian revision.

1. set `OBS-DCH-RELEASE: 1`:
   * :green_circle: deb is VERSION+release_prefix.<CI_CNT>.<B_CNT>
   * :red_circle: rpm is VERSION-release_prefix
2. add `Release: %%{?release_prefix}.<CI_CNT>.<B_CNT>` in prjconf
   * :red_circle: deb is VERSION+release_prefix
   * :green_circle: rpm is VERSION-release_prefix.<CI_CNT>.<B_CNT>

**Details**:

This problem ➜ https://forums.opensuse.org/showthread.php/554631

I believe is caused by this line being too greedy and clearing Release completely:

https://github.com/openSUSE/obs-build/blob/d8665ae4d182527dec454328222491aa86de522b/build-recipe-dsc#L52

ColMelvin:
To handle macros with macros embedded in them, e.g. `%{expand: %%{mymacro}%%{mysuffix}}`, there will need to be a more complicated regex.  An example might be:

```
s/
    (?(DEFINE)
        (?<MACRO> %\{ (?&EXPANSION) \} | %(?&VARIABLE) | %\*\* | %[*#] )
        (?<EXPANSION> [^%]+ | [^%]* (?: (?&MACRO) [^%]* )+ )
        (?<VARIABLE> \w+ )
    )
    (?&MACRO)\.?
//xg;
```

I don't think this solution will work well with parameterized macro calls (e.g `%mymacro 5`), so further iteration may be required.

drmuey:
@ColMelvin and I were able to verify that this is indeed the problem 👍

We get it to work w/ a simplified regex that will only work on the classic prjconf `Release: %%{?release_prefix}.<CI_CNT>.<B_CNT>` (lots of escape issues and special character oddness leyt along complexities of macro format).


First, thi sgot us the correct `DEB_RELEASE`:

```
--- build-recipe-dsc.orig	2021-06-01 12:49:32.606955951 -0500
+++ build-recipe-dsc	2021-06-01 14:42:17.490308842 -0500
@@ -49,7 +49,10 @@
     done
     # remove rpm macros (everything after "%")
     # they are not evaluated by the Debian build process
[...]

PR #698, opened 2021-06-09
Issue 695: Get the standard prjconf release_prefix to work for debs

drmuey:
Standard one in the docs is:
  `Release: %%{?release_prefix}.<CI_CNT>.<B_CNT>`

That is both via `OBS-DCH-RELEASE` and `DEB_TRANSFORM` mechanisms.

Round 2, if anyone wants it, would be to make it smart enough to strip any macros.


drmuey:
#695

PR #700, opened 2021-06-21
Fix Mageia Build

tarakbumba:
- Tests fail in %check phase of build process on Mageia GNU Linux due to missing perl(YAML::PP) package. This commit adds missing build requires. (Also see declined 900291 submit request on OBS)
- Both License and Group tags are incompatible with Mageia packaging system. This commit fixes also this issue

### DIFF
--- a/build-recipe-arch
+++ b/build-recipe-arch
@@ -62,5 +62,5 @@ recipe_cleanup_arch() {
 }
 
 _arch_recipe_makepkg() {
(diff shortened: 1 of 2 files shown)


adrianschroeter:
please rebase and revert the license part, since this is an incorrect change.

adrianschroeter:
this seems to be obsolete

PR #704, opened 2021-06-23
pbuild: fix --shell option

adrianschroeter:
* Make the shell accessable
* No watchdog running

It still needs --rebuild or --singlejob option.


adrianschroeter:
it would be nice just to use last environment when no package is specified, but not in this pull request.

PR #708, opened 2021-07-22
Show constructed rpmlint command line

jengelh:


adrianschroeter:
not really keen on this, at least not without a comment in the code. Since this looks more like a left-over from debugging....

adrianschroeter:
no reaction ... if it all, then via a debug option.

jengelh:
What kind of comment do you want in source? (I hate obvious statements like `# Show rpmlint line`, everybody ought to know what set -x does.)

PR #712, opened 2021-08-03
build-vm: try mounting selinuxfs

dirkmueller:
This appears to be needed for selinux to work. without selinux
enabled, appliance builds seem to fail


bugfinder:
+1 from me

Issue #673, opened 2021-03-05
Dockerfile builds FROM scratch are broken


Vogtinator:
It tries to execute `obs-docker-support` anyway and fails, because there's no `/bin/sh`.

```
[   75s] Step 1/7 : FROM scratch
[   75s]  ---> 
[   75s] Step 2/7 : COPY .obs-docker-support /usr/local/sbin/obs-docker-support
[   75s]  ---> 6af716ec5f6f
[   75s] Step 3/7 : RUN obs-docker-support --install
[   76s]  ---> Running in 3cf2a3dc120c
[   76s] time="2021-03-05T15:28:35.256861607Z" level=info msg="starting signal loop" namespace=moby path=/run/containerd/io.containerd.runtime.v2.task/moby/3cf2a3dc120c30911c9242c087d1685191734a8b66f772ee5dc5815fa26090ef pid=2103
[   76s] time="2021-03-05T15:28:35.598076360Z" level=info msg="shim disconnected" id=3cf2a3dc120c30911c9242c087d1685191734a8b66f772ee5dc5815fa26090ef
[   76s] time="2021-03-05T15:28:35.607494982Z" level=error msg="stream copy error: reading from a closed fifo"
[   76s] time="2021-03-05T15:28:35.607616986Z" level=error msg="stream copy error: reading from a closed fifo"
[   76s] time="2021-03-05T15:28:35.624121107Z" level=error msg="3cf2a3dc120c30911c9242c087d1685191734a8b66f772ee5dc5815fa26090ef cleanup: failed to delete container from containerd: no such container"
[   76s] OCI runtime create failed: container_linux.go:367: starting container process caused: exec: "/bin/sh": stat /bin/sh: no such file or directory: unknown
```

PR #714, opened 2021-08-11
Fix Dockerfile builds FROM scratch

vasiliy-ul:
Do not run obs-docker-support when building from scratch since it will fail anyway because of missing /bin/sh.

Fixes https://github.com/openSUSE/obs-build/issues/673


adrianschroeter:
thanks!

PR #717, opened 2021-08-26
build-recipe-dsc: avoid publishing duplicate dsc files

gollub:
With "OBS-DCH-RELEASE: 1" set two *.dsc files might be published,
which seems to confuse the src_server or scheduler and avoids
correct triggering of rebuilds.

This is most probably an unintended behavior change introduced by
d995de1d71e33.

### DIFF
--- a/build-recipe-dsc
+++ b/build-recipe-dsc
@@ -114,9 +114,9 @@ dsc_build() {
     # this allows the build environment to be manipulated
     # and alternate build commands can be used
     DSC_BUILD_CMD="$(queryconfig --dist "$BUILD_DIST" --archpath "$BUILD_ARCH" --configdir "$CONFIG_DIR" substitute dsc:build_cmd)"
-    test -z "$DSC_BUILD_CMD" && DSC_BUILD_CMD="dpkg-buildpackage -us -uc"
+    test -z "$DSC_BUILD_CMD" && DSC_BUILD_CMD="dpkg-buildpackage -us -uc -b"
     if test -e $buildroot/$TOPDIR/SOURCES/build.script ; then
-	echo "Sourcing build.script to build - it should normally run 'dpkg-buildpackage -us -uc'"
+	echo "Sourcing build.script to build - it should normally run 'dpkg-buildpackage -us -uc -b'"
 	DSC_BUILD_CMD="source $TOPDIR/SOURCES/build.script"
 	chmod +x $buildroot/$TOPDIR/SOURCES/build.script
     fi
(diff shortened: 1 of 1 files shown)


gollub:
With Debian 10 as build target looks like one now might run in this particular lintian issue: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=911864

Closing for now and doing further investigation.

nickbroon:
@gollub @smcv did you ever get to a solution for this?

When `OBS-DCH-RELEASE` is set then `build-recipe-dsc()` is still copying two source packages, since the merge of d995de1.

I think that either the change originally proposed by @gollub in this pull-request, or the one below, is required so that `dsc_move_build_result()` is behaves correctly with `OBS-DCH-RELEASE` in the same way that `dsc_build()` and `recipe_prepare_dsc()` do.

``` diff
--- build-recipe-dsc	2023-04-17 17:42:44.376347313 +0100
+++ build-recipe-dsc.modified	2023-04-17 17:43:35.152369969 +0100
@@ -197,7 +197,7 @@
 	mv "$changes" "$BUILD_ROOT/$TOPDIR/DEBS"
     done

-    if test "$RECIPEFILE" != "debian/control" -a "$RECIPEFILE" != "debian.control" ; then
+    if test "$RECIPEFILE" != "debian/control" -a "$RECIPEFILE" != "debian.control" -a -z "$OBS_DCH_RELEASE" ; then
 	# link used sources over to DEB directory
 	ln $BUILD_ROOT/$DEB_SOURCEDIR/$DEB_DSCFILE $BUILD_ROOT/$TOPDIR/DEBS/
 	while read f ; do
```

smcv:
> did you ever get to a solution for this?

I haven't worked on OBS for months/years, and it's not clear to me exactly what you mean by "this": as I said in an earlier comment, there are several possible workflows for .dsc/.deb packages in OBS, each with its own assumptions.

It would probably be best for issue tracking if you (and anyone else experiencing an issue) opened an issue with steps to reproduce, expected result and actual result, rather than hoping that developers can extrapolate what the issue is from the solution you propose. Replying to a closed merge request is not usually an effective way to make a change happen.

Some factors that are likely to be necessary to mention in an issue report:

* what is your source code? (unpacked source directory? unpacked source directory with .git/? pre-existing .dsc?) - on all the OBS installations I've worked with, our source code has been a pre-existing .dsc, prepared outside OBS with `dpkg-buildpackage --build=source` or equivalent
* configuration: `OBS-DCH-RELEASE` enabled or disabled? - I think this is disabled on all the OBS installations I've worked with
* git-buildpackage as a OBS source-service enabled or disabled? - I think this is disabled on all the OBS installations I've worked with
* what sequence of events do you expect to happen in this scenario? - in the OBS installations I've worked with, what we want is that OBS will unpack the .dsc, bump the version number in the changelog to add the build number, do a `dpkg-buildpackage --build=any,all` (build binaries but don't rebuild the .dsc), and publish the original .dsc and the just-built binaries

nickbroon:
I can open a new ticket track this.

`OBS-DCH-RELEASE` is not an OBS installation setting, its a setting that can be added on per package basis in the .dsc file `OBS-DCH-RELEASE: 1`, so for example can be used on build.opensuse.org. It's simply an option that `build-recipe-dsc` understands.  Wither that comes directly from .dsc file provides as the source, or from a source service like `gbp_scm` that will generated a .dsc (gbp_scm is not available on build.opensuse.org and I don't use it but  itself has a `gbp-dch-release-update` option that will inject `OBS-DCH-RELEASE: 1` into the generated .dsc).

The problem is that by default if `OBS-DCH-RELEASE: 1` is in the .dsc provided as source to `build-recipe-dsc` where-ever that came from then it's default behaviour of calling `dpkg-buildpackage -us -uc` will generate a new .dsc, and it will then copy both of these out as built results.

I'm also unsure which direction is the correct one that `build-recipe-dsc` should follow when the `OBS-DCH-RELEASE ` option is present, either copy our the original .dsc or copy out the generated .dsc. But it should not copy out both. (Prior to d995de1 it would only ever copy out the original .dsc, but as it now parse the .changes it also copies the generated .dsc). For the former, as you mentioned, that might be achieved by having dpkg-buildpackage not even generate a new .dsc file.

PR #718, opened 2021-08-26
Fix Dockerfile builds FROM scratch

vasiliy-ul:
This is a partial revert of 5412727a

Do not abort the build when the base image is not found.

/cc @mlschroe

### DIFF
--- a/build-recipe-docker
+++ b/build-recipe-docker
@@ -56,9 +56,12 @@ recipe_build_docker() {
     touch $BUILD_ROOT/etc/resolv.conf
 
     base_image_path=
+    base_image_tag=$(grep "^\s*FROM" "$RECIPEFILE" | head -n 1 | cut -d" " -f2)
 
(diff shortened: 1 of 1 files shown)


vasiliy-ul:
Hi @adrianschroeter , this PR is a follow-up of the https://github.com/openSUSE/obs-build/pull/714 . Maybe you could also take a look here? Thanks.

build-recipe-docker
```diff
@@ -56,9 +56,12 @@ recipe_build_docker() {
     touch $BUILD_ROOT/etc/resolv.conf
 
     base_image_path=
+    base_image_tag=$(grep "^\s*FROM" "$RECIPEFILE" | head -n 1 | cut -d" " -f2)
```
dirkmueller, line 59: you're checking only the first "FROM" statement, but my understanding is that the first image needs to be non-scratch, only further on layers (multistages) need to be scratch?

dirkmueller: ah, you're having a signle stage with the first stage going from scratch, okay so thats a special case. :+1:

PR #729, opened 2021-12-01
Always include source sub directories

adrianschroeter:
No matter for which build format anymore.

Native OBS can not deal with sub directories, but scm aproaches can.


adrianschroeter:
a version which checks if the directories really belong to git is no submited

PR #730, opened 2021-12-01
build-vm allow overwriting kernel cmdline params

bugfinder:
allow overwriting the currently hardcoded kernel parameters, some kernels fail to load modules on unknonw params:
[   14s] [    1.177944] dracut-pre-udev[124]: modprobe: ERROR: could not insert 'ext4': Unknown symbol in module, or unknown parameter (see dmesg)
[   15s] [    1.247111] ext4: Unknown parameter `allow_unsupported'

### DIFF
--- a/build
+++ b/build
@@ -382,7 +382,9 @@ Known Parameters:
 
   --vm-kernel FILE
   --vm-initrd FILE
+  --vm-cmdline STRING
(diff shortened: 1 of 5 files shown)


bugfinder:
drop, replacing by new instance

PR #736, opened 2021-12-13
build-vm-kvm sandboxing for x86_64

dirkmueller:


adrianschroeter:
fine for me, the only question is do you have verified that also older qemu versions know these parameters (or at least do not fail with them)?

It should be enough when all 2.x versions know them though.

PR #737, opened 2021-12-13
build I/O performance improvements on x86_64

dirkmueller:


build-vm-kvm
```diff
@@ -156,7 +156,9 @@ vm_verify_options_kvm() {
 	    kvm_serial_device=virtio-serial-ccw
 	    kvm_rng_device=virtio-rng-ccw
 	    ;;
-        x86_64)
+	x86_64)
+            kvm_bin="/usr/bin/qemu-system-x86_64"
```
adrianschroeter, line 160: this needs to be verified that the binary is called that way on all the foreign distributions we do still support.
I remember that I had to fiddle regarding this for debian and ubuntu in the past at least ...

dirkmueller: I changed it to test before setting, which should fix this issue (the ubuntu workaround is earlier in the same function)

PR #738, opened 2021-12-14
Rework vm_cmdline handling for kvm case

dirkmueller:
if kernel cmdline parameters are specified, use those
and only fall back to our parameters if not otherwise specified.


bugfinder (approved):
LGTM

PR #741, opened 2021-12-16
Enable tunings also for aarch64 and 32bit x86 builds

dirkmueller:
Remove an outdated check for qemu 2.5 or older - that should
no longer be an issue.


adrianschroeter:
fine with me, but could you also adapt the spec file to conflict with qemu versions below 2.5 ?

dirkmueller:
@adrianschroeter done

PR #747, opened 2021-12-20
Avoid calling rm X just before cp A X

dirkmueller:
cp(1) has a --remove-destination which does that implicitely,
which avoids calling rm and thereby reducing overhead of
execution

### DIFF
--- a/build-pkg-deb
+++ b/build-pkg-deb
@@ -66,8 +66,7 @@ pkg_initdb_deb() {
     deb_setup
     # force dpkg into database to make epoch test work
     if ! test "$BUILD_ROOT/.init_b_cache/rpms/dpkg.deb" -ef "$BUILD_ROOT/.init_b_cache/dpkg.deb" ; then
-	rm -f $BUILD_ROOT/.init_b_cache/dpkg.deb
-	cp $BUILD_ROOT/.init_b_cache/rpms/dpkg.deb $BUILD_ROOT/.init_b_cache/dpkg.deb || cleanup_and_exit 1
+	cp --remove-destination $BUILD_ROOT/.init_b_cache/rpms/dpkg.deb $BUILD_ROOT/.init_b_cache/dpkg.deb || cleanup_and_exit 1
     fi
     DEB_UNSAFE_IO=
     chroot $BUILD_ROOT dpkg --force-unsafe-io --version >/dev/null 2>&1 && DEB_UNSAFE_IO="--force-unsafe-io"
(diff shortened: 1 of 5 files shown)


marcus-h:
On 2021-12-20 00:51:08 -0800, Dirk Mueller wrote:
> cp(1) has a --remove-destination which does that implicitely,
> which avoids calling rm and thereby reducing overhead of
> execution

This is potentially dangerous (unless we have "full control" over
the destination directory). For instance,

> diff --git a/build-vm-zvm b/build-vm-zvm
> index 4d13fe6..04d999d 100644
> --- a/build-vm-zvm
> +++ b/build-vm-zvm
> @@ -378,8 +378,7 @@ vm_fixup_zvm() {
>      vm_initrd="${vm_initrd}-${kernel_version}"
>      # copy initrd to build worker:
>      echo "copy $vm_initrd to $BUILD_ROOT"
> -    rm -rf "${BUILD_ROOT}/boot/${vm_initrd}"
> -    cp -a "${vm_initrd}" "${BUILD_ROOT}/boot"
> +    cp --remove-destination -a "${vm_initrd}" "${BUILD_ROOT}/boot/${vm_initrd}"

${BUILD_ROOT}/boot/${vm_initrd} could be a symlink to a
a directory on the host system. That is, we can write to
a file on the host system...

>      # finally, install bootloader to the worker disk
>      # XXX/TODO: does zipl read or write ${BUILD_ROOT}${vm_initrd}? Either
>      # rm -rf the file or check for a symlink
> diff --git a/init_buildsystem b/init_buildsystem
> index b3db081..4fd600c 100755
> --- a/init_buildsystem
> +++ b/init_buildsystem
> @@ -915,8 +915,7 @@ if test ! -e $BUILD_ROOT/installed-pkg -a ! -e $BUILD_ROOT/.build/init_buildsyst
>         touch $BUILD_ROOT/etc/fstab
>      fi
>      if test ! -e $BUILD_ROOT/etc/ld.so.conf -a -f $BUILD_ROOT/etc/ld.so.conf.in -a ! -L $BUILD_ROOT/etc/ld.so.conf.in ; then
> -       rm -rf $BUILD_ROOT/etc/ld.so.conf
> -       cp $BUILD_ROOT/etc/ld.so.conf.in $BUILD_ROOT/etc/ld.so.conf
> +       cp --remove-destination $BUILD_ROOT/etc/ld.so.conf.in $BUILD_ROOT/etc/ld.so.conf

Hmm this should be ok: if $BUILD_ROOT/etc/ld.so.conf is a symlink to
an existing directory, the if-branch not executed. Otherwise, the
dangling symlink is removed by cp.

PR #749, opened 2021-12-20
Avoid calling external commands in a loop

dirkmueller:
Batch calling external commands in a loop
    
In many cases rather than invoking the same command in a loop
we can collect all the arguments and use xargs to batch up the
invocations. this reduces the per-loop iteration overhead.
    
In the smallest package in tumbleweed this saves about 1s
of execution overhead on a very fast x86_64 cpu, and 3-5s on
a slower aarch64 cpu. larger packages that install more dependencies
have better outcomes.
    
measurement of just init_buildsystem for xterm package setup on
a tmpfs store:
    
before:
    
      2.63user 1.02system 0:03.32elapsed 109%CPU (0avgtext+0avgdata 36232maxresident)k
      584inputs+496288outputs (9major+576786minor)pagefaults 0swaps
    
after:
    
      1.81user 0.74system 0:02.26elapsed 112%CPU (0avgtext+0avgdata 36052maxresident)k
      584inputs+496288outputs (20major+246016minor)pagefaults 0swaps

### DIFF
--- a/dist/build.spec
+++ b/dist/build.spec
@@ -37,6 +37,7 @@ BuildArch:      noarch
 # Keep the following dependencies in sync with obs-worker package
 Requires:       bash
 Requires:       binutils
+Requires:       findutils
 Requires:       perl
 Requires:       tar
 # needed for fuser
--- a/dist/debian.control
+++ b/dist/debian.control
@@ -7,12 +7,11 @@ Standards-Version: 3.7.2
 
 Package: obs-build
 Architecture: all
-Depends: ${perl:Depends}, rpm, libarchive-tools | bsdtar
+Depends: ${perl:Depends}, rpm, findutils, libarchive-tools | bsdtar
 Recommends: rpm2cpio
 Conflicts: build
 Replaces: build
 Provides: build
-Description: A script to build SUSE Linux RPMs
- This package provides a script for building RPMs for SUSE Linux
+Description: A script to build SUSE RPMs
+ This package provides a script for building RPMs for SUSE
  in a chroot environment.
-
--- a/init_buildsystem
+++ b/init_buildsystem
@@ -221,10 +221,28 @@ preinstall_image_filter() {
     done
 }
 
+# call with cmd [OPTIONS] -- ARGS...
+run_cmd_with_xargs()
+{
+    local cmd=()
+    while [ "${#@}" -gt 1 ]; do
+	cmd[${#cmd[@]}]="$1"
+        shift
+        if [ "$1" = "--" ]; then
+            shift
+            break
+        fi
+    done
+    for i in "$@"; do
+        printf "%s\0" "$i"
+    done | xargs -0 --no-run-if-empty "${cmd[@]}" --
+}
+
 preinstall_setup() {
     cd "$BUILD_ROOT" || cleanup_and_exit 1
     rm -rf build-preinstall
     mkdir build-preinstall || cleanup_and_exit 1
+    local mv_args=()
     for i in * .* ; do
 	# please keep in sync with recipe_build_preinstallimage
 	case "$i" in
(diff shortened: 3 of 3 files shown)


marcus-h:
On 2021-12-20 14:05:28 +0000, Dirk Mueller wrote:
> diff --git a/build-pkg-rpm b/build-pkg-rpm
> index b630bac5..727b23f8 100644
> --- a/build-pkg-rpm
> +++ b/build-pkg-rpm
> @@ -158,16 +158,17 @@ pkg_finalize_rpm() {
>  	echo "now installing cumulated packages"
>  	for ((num=0; num<=cumulate; num++)) ; do
>  	    echo ${CUMULATED_LIST[$num]}
> +	done > $BUILD_ROOT/.init_b_cache/manifest

Is there a reason why you moved it into a separate loop? (I'm fine
with two loops - I'm just interesting if I overlook something:) )

> +	for ((num=0; num<=cumulate; num++)) ; do
>  	    PKG=${CUMULATED_LIST[$num]##*/}
>  	    test "$BUILD_ROOT/.init_b_cache/rpms/$PKG" -ef "$BUILD_ROOT/${CUMULATED_LIST[$num]}" && continue
> -	    rm -f $BUILD_ROOT/${CUMULATED_LIST[$num]}
> -	    cp $BUILD_ROOT/.init_b_cache/rpms/$PKG $BUILD_ROOT/${CUMULATED_LIST[$num]} || cleanup_and_exit 1
> -	done > $BUILD_ROOT/.init_b_cache/manifest
> -	( cd $BUILD_ROOT && chroot $BUILD_ROOT rpm --ignorearch --nodeps -Uh --oldpackage --ignoresize --verbose $RPMCHECKOPTS \
> +	    cp --remove-destination $BUILD_ROOT/.init_b_cache/rpms/$PKG $BUILD_ROOT/${CUMULATED_LIST[$num]} || cleanup_and_exit 1
> +	done
> +	( cd $BUILD_ROOT && chroot $BUILD_ROOT rpm --ignorearch --nodeps -Uvh --oldpackage --ignoresize $RPMCHECKOPTS \
>  		$ADDITIONAL_PARAMS .init_b_cache/manifest 2>&1 || touch $BUILD_ROOT/exit )
>  	for ((num=0; num<=cumulate; num++)) ; do
> -	    rm -f $BUILD_ROOT/${CUMULATED_LIST[$num]}
> -	done
> +            printf "%s\0" "$BUILD_ROOT/${CUMULATED_LIST[$num]}"
> +        done | xargs -0 -r rm -f --
>  	rm -f $BUILD_ROOT/.init_b_cache/manifest
>  	check_exit
>  	for ((num=0; num<=cumulate; num++)) ; do


> diff --git a/init_buildsystem b/init_buildsystem
> index 4fd600c1..dccb9488 100755
> --- a/init_buildsystem
> +++ b/init_buildsystem
> @@ -935,12 +932,14 @@ fi
>
>  if test -n "$PREPARE_VM" ; then
>      echo "copying packages..."
> -    for PKG in $PACKAGES_TO_ALL $PACKAGES_TO_SYSROOTINSTALL ; do
> -	rm -rf $BUILD_ROOT/.init_b_cache/$PKG.$PSUF
> -	cp $BUILD_ROOT/.init_b_cache/rpms/$PKG.$PSUF $BUILD_ROOT/.init_b_cache/$PKG.$PSUF || cleanup_and_exit 1
> -	ln -s -f ../$PKG.$PSUF $BUILD_ROOT/.init_b_cache/rpms/$PKG.$PSUF
> -	check_exit
> -    done
> +
> +    PKGS="$PACKAGES_TO_ALL $PACKAGES_TO_SYSROOTINSTALL"
> +    rm -f $(for P in $PKGS; do echo $BUILD_ROOT/.init_b_cache/$P.$PSUF; done)

Hmm couldn't this result in E2BIG if we have "lots of" packages?
(on my system _SC_ARG_MAX is 2MiB and assuming that the size of a filename
is upper bounded by 255 bytes, we have "lots of" is roughly 4112 packages
(1024*1024 / 255)).
Maybe we can pipe it through xargs?

> +    for P in $PKGS; do printf "%s\0" "$BUILD_ROOT/.init_b_cache/rpms/$P.$PSUF"; done | \
> +        xargs -0 -r cp -L -p -t $BUILD_ROOT/.init_b_cache -- || cleanup_and_exit 1

Why the --dereference (== -L)?
(Why the -p?)
Couldn't we add "--remove-destination" to the cp call in order to get rid
the previous rm call (just an idea: I haven't checked it)?

> +    ln -s -f $(for P in $PKGS ; do echo "../$P.$PSUF"; done) -t $BUILD_ROOT/.init_b_cache/rpms

This could result in E2BIG.


Otherwise, nice! LGTM!

dirkmueller:
> Is there a reason why you moved it into a separate loop? (I'm fine
> with two loops - I'm just interesting if I overlook something:) )

I found it dangerous that we execute an for...; echo ... ; done > manifest with more than just echo in the for loop (other things can produce output on stdout which ends up garbling manifest file). No other reason. it is an unrelated cleanup so we can definitely put it in a different commit. just happened to be in the same hunk. 

> Hmm couldn't this result in E2BIG if we have "lots of" packages?

yes. I think in realistic states we will get this around 32000 packages which I don't believe we'll get soon. but you're right, it could happen, and we can use xargs(1) here as well. I have changed it. 


> Why the --dereference (== -L)?
> (Why the -p?)

I was playing with some other optimisation that avoids an extra copy in cache setup, but it wasn't ready. removed this again for now. 

thank you for the review!

PR #755, opened 2022-01-10
Get vm-type=qemu working again

adrianschroeter:


adrianschroeter:
@dirkmueller this is touching some places you touched recently, can you verify these commits?

build-vm
```diff
@@ -101,6 +102,7 @@ vm_detach_swap() {
 }
 
 vm_fixup() {
+    test -z "$VM_DEBUG" && export vm_linux_always_append="$vm_linux_always_append quiet"
```
dirkmueller, line 105: is the export needed here?

build-vm-qemu
```diff
@@ -212,9 +212,8 @@ vm_kill_qemu() {
 
 vm_fixup_qemu() {
     vm_fixup_kvm
-
-    case $qemu_device in
-        virtio*)
+    case $BUILD_HOST_ARCH in
+        armv6l|armv7l|armv8l|aarch32|aarch64|aarch64_ilp32|ppc|ppcle|ppc64|ppc64le|riscv64|s390|s390x)
```
dirkmueller, line 216: why is this not including hppa and the other architectures where `vm_startup_qemu()` is setting `qemu_device` ? what is the reason for changing away from that detection to the build host? shouldn't the setting of qemu_device then removed as this is the only place where it was referenced?

build-vm-qemu
```diff
@@ -224,13 +223,14 @@ vm_fixup_qemu() {
             ;;
     esac
 
-    if test -z "$qemu_serial_device" ; then
-        if test -e "$BUILD_ROOT/.build.console.kvm" -a ! -L "$BUILD_ROOT/.build.console.kvm" && grep -q '^virtio$' "$BUILD_ROOT/.build.console.kvm" ; then
```
dirkmueller, line 228: I think a slightly better fix is:

```patch
-        if test -e "$BUILD_ROOT/.build.console.kvm" -a ! -L "$BUILD_ROOT/.build.console.kvm" && grep -q '^virtio$' "$BUILD_ROOT/.build.console.kvm" ; then
+        if test -e "$BUILD_ROOT/.build.console.qemu" -a ! -L "$BUILD_ROOT/.build.console.qemu" && grep -q '^virtio$' "$BUILD_ROOT/.build.console.qemu" ; then
```

build-vm-qemu
```diff
@@ -224,13 +223,14 @@ vm_fixup_qemu() {
...
-            echo "Detected virtio-serial support"
-            qemu_serial_device=virtio-serial,max_ports=2
-            qemu_console=hvc0
-        fi
-    fi
+# virtio serial built-in seems not to work since qemu 5.2 anymore
```
dirkmueller, line 226: it works just fine here? I think the problem you're facing is that `vm_startup_qemu()` is overriding `qemu_console` with ttyAMA0 rather than keeping `qemu_console` at hvc0 like set here.

the underlying issue is that the bulk of the qemu* device determination logic needs to be executed much earlier, like in the kvm case it is done in `vm_verify_options_kvm` which however is too early for qmeu as the `$BUILD_HOST_ARCH` isn't set at that point. 

I see two options: move determination of BUILD_HOST_ARCH much earlier and move the device determination logic into `vm_verify_options_qemu`. the other option I see is that we share move about the qemu_storage device and drive as well as the virtio serial determination logic with the kvm case. 

by a quick hack I did a sed /qemu_device/kvm_device/ on build-vm-kvm and that makes it work for me. obviously thats not a nice fix but the virtio detection and serial detection logic all works naturally then.

build-vm
```diff
@@ -294,6 +296,9 @@ vm_parse_options() {
         VM_OPENSTACK_FLAVOR="$ARG"
         shift
       ;;
+      -vm-debug)
```
dirkmueller, line 299: since its just removing the quiet, can we call it -vm-verbose maybe?

build-vm-qemu
dirkmueller: answering my own question, I see the problem is that at the time of execution of `vm_fixup_qemu()` the `$qemu_device` isn't set already based on BUILD_HOST_ARCH, as it is only set at vm_startup_qemu times, where it is much too late and breaking all the other subtle logic in build-vm-km.

adrianschroeter:
On Mittwoch, 12. Januar 2022, 00:48:12 CET Dirk Mueller wrote:
> @dirkmueller commented on this pull request.
> 
> 
> 
> > @@ -294,6 +296,9 @@ vm_parse_options() {
>          VM_OPENSTACK_FLAVOR="$ARG"
>          shift
>        ;;
> +      -vm-debug)
> 
> since its just removing the quiet, can we call it -vm-verbose maybe?

yes, maybe a better name, since --debug exists as well (which has a complete
different intention, building debug packages, but not for debugging environment ...)

will change that.

-- 

Adrian Schroeter ***@***.***>

SUSE Software Solutions Germany GmbH, Maxfeldstr. 5, 90409 Nürnberg, Germany
(HRB 36809, AG Nürnberg) Geschäftsführer: Ivo Totev

build-vm-qemu
adrianschroeter: it was indeed the kernel parameter. I changed the code now to avoid overwriting an already set console name.

dirkmueller (approved):

adrianschroeter:
urgs, pushed it directly by accident, but should be fine

PR #757, opened 2022-01-12
docker: support package installation calls inside of if conditions

adrianschroeter:
Always parse them in if _and_ else conditions as we can not judge
about the runtime check


Vogtinator:
There's also `do`. And `if` might be useful as well, for cases like `if zypper ...`

Issue #500, opened 2019-05-01
Support building debian packages that use dh-exec


nickbroon:
If the debian.tar.gz contains an executable file, as is needed if using dh-exec as part of the packaging, the build fails when using OBS.
Example: https://build.opensuse.org/package/show/home:nickbrown:branches:openSUSE:Tools/osc

I see:

```
[  229s] make[1]: Leaving directory '/usr/src/packages/BUILD'
[  229s]    dh_install
[  229s] cp: cannot stat 'debian/tmp/=>': No such file or directory
[  229s] dh_install: cp -a debian/tmp/=> debian/osc//usr/lib/osc/complete/ returned exit code 1
[  229s] debian/rules:4: recipe for target 'binary' failed
```
and my osc.install contains, and is marked executable in the debian.tar.gz:
```
#!/usr/bin/dh-exec
dist/osc.complete => /usr/lib/osc/complete
```
But `debtransform` turn this into .diff.gz format 1.0 file, and osc.install is no longer executable, resulting in the above error.

dh-exec is a widely used debian packaging helper. https://packages.debian.org/source/stretch/dh-exec

noscript:
Experiencing the same problem. As a workaround I'm renaming inside `override_dh_auto_install`.

Issue #369, opened 2017-07-16
[Debian] Support for newer source formats is needed


zilti:
See below; support for newer source formats is needed. Seeing as there is another bug asking about that that is way over a year old now I also have to ask if it's even planned at all to keep supporting `.deb` building?

```
[   83s]  dpkg-source -b BUILD
[   83s] dpkg-source: warning: no source format specified in debian/source/format, see dpkg-source(1)
[   83s] dpkg-source: warning: source directory 'BUILD' is not <sourcepackage>-<upstreamversion> 'chicken-4.12.0'
[   83s] dpkg-source: warning: .orig directory name BUILD.orig is not <package>-<upstreamversion> (wanted chicken-4.12.0.orig)
[   83s] dpkg-source: info: using source format '1.0'
[   83s] dpkg-source: info: building chicken using existing chicken_4.12.0.orig.tar.gz
[   84s] dpkg-source: info: building chicken in chicken_4.12.0-22.1.diff.gz
[   84s] dpkg-source: warning: newly created empty file 'debian/changelog.dpkg-orig' will not be represented in diff
[   84s] dpkg-source: warning: newly created empty file 'debian/compat.dpkg-orig' will not be represented in diff
[   84s] dpkg-source: warning: newly created empty file 'debian/control.dpkg-orig' will not be represented in diff
[   84s] dpkg-source: warning: newly created empty file 'debian/rules.dpkg-orig' will not be represented in diff
[   84s] dpkg-source: error: diff '/usr/src/packages/chicken_4.12.0-22.1.diff.gz.new.Zgi2bP' patches file with name ending in .dpkg-orig
[   84s] dpkg-buildpackage: error: dpkg-source -b BUILD gave error exit status 29
```

bluca:
obs-build does support format 3.0 just fine, eg:

```
[   80s] -----------------------------------------------------------------
[   80s] ----- building dpdk_16.11.3-0vyatta1.dsc (user abuild)
[   80s] -----------------------------------------------------------------
[   80s] -----------------------------------------------------------------
[   81s] dpkg-buildpackage: info: source package dpdk
[   81s] dpkg-buildpackage: info: source version 16.11.3-0vyatta1+18.2
[   81s] dpkg-buildpackage: info: source distribution unstable
[   81s] dpkg-buildpackage: info: source changed by Luca Boccassi <lboccass@brocade.com>
[   81s]  dpkg-source --before-build BUILD
[...]

Issue #225, opened 2015-11-30
Cannot build Debian/Ubuntu packages that require binaries added to source tree


delcypher:
Unfortunately I'm building a package where a small binary file needs to be added to the source tree before configuring and building

The way I achieved getting the binary there is by adding the following tag to my ``.dsc`` file.

```
DEBTRANSFORM-FILES: my_binary_file
```

when I try to build it fails

```
[   85s] ** Started: debtransform //usr/src/packages/SOURCES //usr/src/packages/SOURCES/z3.dsc //usr/src/packages/SOURCES.DEB                                                   [17/1995]
[   85s] No DEBTRANSFORM-FILES-TAR line in the .dsc file.
[   85s] Attempting automatic discovery of a debian archive.
[   85s] Hardlinking //usr/src/packages/SOURCES/z3-4.4.1.tar.gz to //usr/src/packages/SOURCES.DEB/z3_4.4.1.orig.tar.gz
[   85s] files 4336a9df24f090e711c6d42fd4e2b1fc 3347371 z3_4.4.1.orig.tar.gz
[   85s] Scanning //usr/src/packages/SOURCES.DEB/z3_4.4.1.orig.tar.gz...
[   86s] Generating //usr/src/packages/SOURCES.DEB/z3_4.4.1-0.diff
[   86s] Processing file "//usr/src/packages/SOURCES/Microsoft.Z3.mono.snk"...
[   86s] Processing file "//usr/src/packages/SOURCES/debian.rules"...
[   86s] Processing file "//usr/src/packages/SOURCES/debian.control"...
[   86s] Processing file "//usr/src/packages/SOURCES/debian.changelog"...
[   86s] Writing //usr/src/packages/SOURCES.DEB/z3_4.4.1-0.dsc
[   86s] dpkg-source: warning: extracting unsigned source package (/usr/src/packages/SOURCES.DEB/z3_4.4.1-0.dsc)
[   86s] dpkg-source: info: extracting z3 in /usr/src/packages/BUILD
[   86s] dpkg-source: info: unpacking z3_4.4.1.orig.tar.gz
[   86s] dpkg-source: info: applying z3_4.4.1-0.diff.gz
[   86s] dpkg-source: info: upstream files that have been modified: 
[   86s]  /usr/src/packages/BUILD/Microsoft.Z3.mono.snk
[   86s] -----------------------------------------------------------------
[   86s] ----- building z3.dsc (user abuild)
[   86s] -----------------------------------------------------------------
[...]

PR #760, opened 2022-01-18
Add support of Debian Source format 3.0 (quilt)

ila-embsys:
This merge request contains changes to get a possibility to use Debian source format version 3.0 (quilt).

[Debian Manual](https://www.debian.org/doc/debian-policy/ch-controlfields.html#id26) describe supported format list:
"1.0, 3.0 (native), and 3.0 (quilt)". To pick differences, as the reference was used [the page](https://wiki.debian.org/Projects/DebSrc3.0) which summarizes how to switch source package formats.

Using this information, changes in the merge request is bringing to OBS's Debian source generator next:
 - Supports inclusion of binary files. It is done by the existing `Debtransform-Files` tag by keeping them in `*.debian.tar.gz` and passing `--include-binaries` to `dpkg-source`.
 - Debian-specific changes are no longer stored in a single `.diff.gz` but in multiple patches in `debian/patches/`. inside `*.debian.tar.gz`. Previously a file `debian.series` was used to make `.diff.gz`.

Key changes were done in `debtransform` file.
To keep the previous behaviour for 1.0, the script checks in necessary places format version in `*.dsc` and apply functions that are changed to specific for generating the new source format. Ex: Functions `doseries_quilt`, `dofile_quilt`, `dotar_quilt` for new format is used rather `doseries`, `dofile`, `dotar`.

Into the folder `test-debtransform`, similar to existing tests added the new test that was used to check the `debtransform` generator with the new source format.


ila-embsys:
Resolves #225. Binary files were one of the reasons to start the work.
Possibly resolves #369. The merge request brings the new format but only `3.0 (quilt)`. It will not affect `3.0 (native)`.
Possibly resolves #500. Modified code that in this merge request was used to build Debian package with `install` Make target. An error with `dh_install` was not noticed.

ila-embsys:
> IIRC recently support was added to avoid needing the .dsc file in the sources (given it's like the srpm, it's a generated output not an input), any chance to support that? The source format is recorded in the debian.tar in debian/source/format

I am, not sure I understand you. Maybe I provide not entirely accurate description.

There are two `*.dsc` files.

The first one is an input for the `debtransform`. It contains sections like 'DEBTRANSFORM-TAR' and 'DEBTRANSFORM-FILES'. I will call it a `dsc template`.
The second one is the output of the `debtransform`. It contains the section `Files` with references to a generated `*.orig.tar.gz` and `*.debian.tar.gz` archives. It is a true `*.dsc`. I will call it the `true dsc`.

So, a `dsc template` is passed to the `debtransform`. If the `debtransform` sees that in the `dsc template` the `Format` field has the value `3.0 (quilt)` then the `debtransform` generate three files: an `*.orig.tar.gz`, a `*.debian.tar.gz` and the `true dsc`. The `true dsc` is outside of generated sources. Also in the file `*.debian.tar.gz` in the path `debian/source/format` the `debtransform` writes the `3.0 (quilt)` value.

Is this the behaviour you have asked me about?

bluca:
OBS can now work (AFAIK) without a .dsc in the source, as long as the standard debian/control file is available - that's what I'm referring to

ila-embsys:
> OBS can now work (AFAIK) without a .dsc in the source, as long as the standard debian/control file is available - that's what I'm referring to

[Here is original repository](https://build.opensuse.org/package/show/home:ila.embsys/zephyr-sdk-ada) that I used to build the package when I worked on this merge request. How can I build it without a `dsc`? Even though there is `debian.control`.

This is my first attempt to package `deb` by OBS. I read the Debian Builds article section [Alternative method of creating a Debian package](https://en.opensuse.org/openSUSE:Build_Service_Debian_builds#openSUSE_prerequisites_to_build_.deb_packages). Have I miss some important? I gather sources by `obs_scm` then generate debian-tar + dsc by `debtransform` then build the package itself. I see only a way of prepare debian-tar manually by hand but it is not what I want.

bluca:
Simply omit the .dsc file

ila-embsys:
> Simply omit the .dsc file

I tried just omit `*.dsc` on remote repo. You are right, OBS is trying to build the package using `debian.control` file as a recipe.

Unfortunately, the build process failed because of `debian/changelog` is not found. Where should it be? On the root of the repo or on `*.debian.tar.gz`? I have tried to build the package in the same way locally without `*.dsc`. But my local runner didn't build the package with `debian.control` file as a recipe.

Can you provide more information on how to use the build without `*.dsc`?

bluca:
> > Simply omit the .dsc file
> 
> I tried just omit `*.dsc` on remote repo. You are right, OBS is trying to build the package using `debian.control` file as a recipe.
> 
> Unfortunately, the build process failed because of `debian/changelog` is not found. Where should it be? On the root of the repo or on `*.debian.tar.gz`? I have tried to build the package in the same way locally without `*.dsc`. But my local runner didn't build the package with `debian.control` file as a recipe.
> 
> Can you provide more information on how to use the build without `*.dsc`? Is it documented anywhere?

That expects it to be in the root of the repository - ie, a standard Debian package repository layout.

But anyway, it was not my intention to block with PR - this is a great starting point, this feature can be improved later.

bluca (approved):

ila-embsys:
> That expects it to be in the root of the repository - ie, a standard Debian package repository layout.
> 
> But anyway, it was not my intention to block with PR - this is a great starting point, this feature can be improved later.

I understood the idea. This is less related to the `debtransform` tool, and sure, can be improved later.

Thank you for the review. :)

ila-embsys:
@adrianschroeter Can you look the request again? It is approved now.

PR #762, opened 2022-01-19
support OBS-BuildName in kiwi files

adrianschroeter:
Support also @BUILD_FLAVOR@ replacement for kiwi and docker build result name


adrianschroeter:
broken approach, need to work based on profiles ...

PR #763, opened 2022-01-19
KIWI: Allow to extend cointainer name with profile

adrianschroeter:
This should be default, but changing the default would break existing
aggregates.

This has also a limitation that it only works properly when build just
one profile per build atm.

### DIFF
--- a/Build/Kiwi.pm
+++ b/Build/Kiwi.pm
@@ -246,6 +246,7 @@ sub kiwiparse {
     $obsprofiles = [ grep {defined($_)} map {$_ eq '@BUILD_FLAVOR@' ? $buildflavor : $_} split(' ', $obsprofiles) ];
   }
   $unorderedrepos = 1 if $xml =~ /^\s*<!--\s+OBS-UnorderedRepos\s+-->\s*$/im;
+  $ret->{'appendprofiletocontainername'} = 1 if $xml =~ /^\s*<!--\s+OBS-AppendProfileToContainername\s+-->\s*$/im;
   for ($xml =~ /^\s*<!--\s+OBS-Imagerepo:\s+(.*)\s+-->\s*$/img) {
     push @imagerepos, { 'url' => $_ };
   }
@@ -625,6 +626,10 @@ sub showcontainerinfo {
     'buildtime' => $buildtime,
     '_type' => {'buildtime' => 'number'},
   };
+  if ($d->{'appendprofiletocontainername'} && @{$d->{'profiles'} || []}) {
+    # we can currently only handle one profile
+    $containerinfo->{'name'} .= "-$d->{'profiles'}->[0]";
+  }
   $containerinfo->{'version'} = $d->{'version'} if defined $d->{'version'};
   $containerinfo->{'release'} = $release if defined $release;
   $containerinfo->{'tags'} = $d->{'container_tags'} if @{$d->{'container_tags'} || []};


mlschroe:
No, that's not possible.

mlschroe:
Let me explain: of course it's possible to use the repository of the first tag if the tag-to-add starts with ':'. But you still want a different container name, so that both the containers are visible in the :full tree.

mlschroe:
(The tags are like rpm provides.)

Vogtinator:
As only aggregates are affected, can we have this as prjconf flag instead? That way we don't have to uglify kiwi descriptions until the end of time.

mlschroe:
I wrote the script that checked all of the existing aggregates, and the only affected packages are:

IBS
---
home:favogt:containers-sle15/15.0
home:favogt:containers-sle15/15.1
home:favogt:releasetest:Containers:PrjS/15.0
home:favogt:releasetest:Containers:PrjS/15.1
home:msmeissn/containers
SUSE:Containers:SLE-SERVER:15/15_0
SUSE:Containers:SLE-SERVER:15/15_1
SUSE:Containers:SLE-SERVER:15/15_2
SUSE:Containers:SLE-SERVER:15/15_3
SUSE:Containers:SLE-SERVER:15/15_4
SUSE:Containers:SLE-SERVER:15/kubevirt_15_2
SUSE:Containers:SLE-SERVER:15/kubevirt_15_3
SUSE:Containers:SLE-SERVER:15/kubevirt_15_4
SUSE:Containers:SLE-SERVER:15/others


OBS
---
home:favogt:releasetest:Leap/15.0
home:favogt:releasetest:Leap/15.0-ports
home:favogt:releasetest:Leap/15.3-dnf
home:favogt:releasetest:Leap/42.3
openSUSE:Containers:Leap/15.0
openSUSE:Containers:Leap/15.0-ports
openSUSE:Containers:Leap/15.1
openSUSE:Containers:Leap/15.1-ppc64le
openSUSE:Containers:Leap/15.2
openSUSE:Containers:Leap/15.2.1
openSUSE:Containers:Leap/15.2-ppc64le
openSUSE:Containers:Leap/15.3
openSUSE:Containers:Leap/15.3-armv7l
openSUSE:Containers:Leap/15.3-dnf
openSUSE:Containers:Leap/15.4
openSUSE:Containers:Leap/15.4-armv7l
openSUSE:Containers:Leap/15.4-dnf
openSUSE:Containers:Leap/42.3

(The OBS scan is still in progress...)

Almost all of the above do not use any profile, so they are safe as well.

Thus, I changed my mind and now consider it a severe bug that the profile is not in the container name. I.e. we should change the code to always add it if a profile is used.

Vogtinator:
Yay!

On IBS, I'm not aware of any container built with kiwi profiles + multibuild.

On OBS, Leap and Tumbleweed containers use them, but only Leap passes through `_aggregates`. It'll only affect 15.3 and 15.4 though, the others aren't built anymore and will keep working.

Only slightly tricky part is the transition time. The `_aggregate`s need to work with both current and new OBS behaviour until the change is completed. For 15.3, there will be images with the old name staying around, so those have to be compatible anyway. That should be easy to implement by just aggregating the old and new names:

```
<aggregatelist>
  <aggregate project="openSUSE:Containers:Leap:15.3">
    <binary>container:opensuse-leap-image</binary>
    <binary>container:opensuse-leap-image-docker</binary>
  </aggregate>
</aggregatelist>
```

Vogtinator:
I adjusted `openSUSE:Containers:Leap/15.3{,-dnf}` and `openSUSE:Containers:Leap/15.4{,-dnf}` to list both names.

PR #766, opened 2022-01-22
Support OBS Debuginfo build flag for Red Hat variants

kailiu42:
By undefining %_enable_debug_packages , debuginfo package will not be
created by rpmbuild on Red Hat variant Linux.

Signed-off-by: Kai Liu <kai.liu@suse.com>

### DIFF
--- a/build-recipe-spec
+++ b/build-recipe-spec
@@ -184,6 +184,11 @@ recipe_build_spec() {
 	rpmbopts[${#rpmbopts[@]}]='--define'
 	rpmbopts[${#rpmbopts[@]}]="_build_create_debug 1"
     fi
+    # OBS Debuginfo flag support for Red Hat family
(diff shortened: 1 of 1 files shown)


mlschroe:
Shouldn't the undefine of '_enable_debug_packages' be enough? I'm a bit reluctant to undefine the debug_package macros for all distributions.

kailiu42:
In rpmpopt:

```
rpmbuild alias --nodebuginfo    --define 'debug_package %{nil}' \
        --POPTdesc=$"do not generate debuginfo for this package"
```

That's why I have the `debug_package` set. But yes, undefining `_enable_debug_packages` is enough to stop debug{info,source} packages from been generated.

kailiu42:
Removed `--define 'debug_package %{nil}'`.

mlschroe:
Thanks!

PR #770, opened 2022-01-25
Fix macro %if statements

adrianschroeter:
it broke parsing, since %if got replaced by macro definition, leading
to break all other %if statements in build description

github issue 12127


adrianschroeter:
https://github.com/openSUSE/obs-build/issues/769

adrianschroeter:
rpm is not allowing it either as macro

PR #772, opened 2022-01-30
avoid printing transfer output

dirkmueller:
But still show all the errors like running out of disk space
and the like.

### DIFF
--- a/build-vm
+++ b/build-vm
@@ -358,7 +358,7 @@ vm_img_create() {
 
     mkdir -p "${img%/*}" || cleanup_and_exit 4
     # truncate file to the desired size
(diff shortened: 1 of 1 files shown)


lnussel:
With this applied we still get random dd output. We really need status=none.

```
$ dd if=/dev/zero of=/dev/null count=1
1+0 records in
1+0 records out
512 bytes copied, 0.000212356 s, 2.4 MB/s
$ dd if=/dev/zero of=/dev/null count=1 status=noxfer
1+0 records in
1+0 records out
$ dd if=/dev/zero of=/dev/null count=1 status=none
$ 
```

PR #773, opened 2022-01-30
support preinstallimages in zst format

dirkmueller:
Preinstallimages are intended to speed up package setup, but currently
they no longer do that as they're compressed by gzip, while (tumbleweed)
uses zstd, which decompresses faster.

So switch to zstd which allows decompression speeds that are factor 3-4
higher (allowing ~ 700MB/s of decompression speed on not too old
x86_64).


adrianschroeter:
problem could be that the systems which are doing the unpacking not necessarly have zstd or tar/bsdtar with enabled zstd support. So it could be seen as incompatible change. 

However, we could catch this and report it as badhost with error message at least.

In any case, I would implement using an external decompressor, instead of depending on enabled support in tar/bsdtar in init_buildsystem preinstall_image()

adrianschroeter:
I think this will also affect remote instances, so we need to release also a build script for OBS 2.10

adrianschroeter:
remote instances are not using preinstallimages, but osc may. we need to check this ...

dirkmueller:
@adrianschroeter  my understanding is that when the preinstallimage can not be unpacked, it should simply fall back to "traditional" install. 

I can see the following options:

* it becomes a buildflag, so that only e.g. tumbleweed has the new preinstallimages, others have old ones
* I change the PR to always also create the tar.gz for fallback with older distros (downside ~ twice the download for everyone)
* I ensure that everything fallsback to non-preinstall image when the preinstall failed for whatever reason. 

opinions?

adrianschroeter:
there is a fallback when the image does not exist. But when it exists, it must be extractable. The build script would not hand back to the tooling (osc/bs_worker).

but we could make it configurable indeed, either via prjconf and/or via the preinstallimage build description.

dirkmueller:
@adrianschroeter  added preinstallimagetype which can be set to the value "zstd" to enable the feature. defaults to gzip builds. 

compressing in two steps as you required actually makes the compression even better because zstd can look at the full file rathre than just the 4k lookahead it gets from the pipe/tar stream.

build-recipe-preinstallimage
```diff
@@ -57,10 +57,21 @@ recipe_build_preinstallimage() {
...
     fi
     echo "image created."
+    local compress=gzip
+    local ext=gz
+    preinstallimagetype=$(queryconfig --dist "$BUILD_DIST" --configdir "$CONFIG_DIR" --archpath "$BUILD_ARCH" buildflags preinstallimagetype)
+    if test "$preinstallimagetype" = "zstd" -a -n "$(type -p zstd)"; then
+        compress="zstd -19 -T0"
+        ext=zst
+    else
+        echo "Using legacy gzip preinstall images."
```
mlschroe, line 71: I don't like the word "legacy". Just do "Using '$compress' for image compression".

build-recipe-preinstallimage
```diff
@@ -70,7 +81,7 @@ recipe_build_preinstallimage() {
 	test -n "$PKG_HDRMD5" || cleanup_and_exit 1
 	echo "$PKG_HDRMD5  $PKG" >> $BUILD_ROOT$TOPDIR/OTHER/preinstallimage.info
     done
-    mv $BUILD_ROOT/.preinstallimage.$$.tar.gz $BUILD_ROOT$TOPDIR/OTHER/preinstallimage.tar.gz
+    mv $BUILD_ROOT/.preinstallimage.$$.tar.$ext $BUILD_ROOT$TOPDIR/OTHER/preinstallimage.tar.$ext || cleanup_and_exit 1
```
mlschroe, line 84: No need for $ext, just use `*`

mlschroe (changes requested):

build-recipe-preinstallimage
dirkmueller: that doesn't work because for some reason that I don't understand it is being renamed from `.preinstallinstallimage.$$.tar.$ext` to `preinstallimage.tar.$ext` with this move.

build-recipe-preinstallimage
dirkmueller: done

build-recipe-preinstallimage
dirkmueller: seems like it has been there from day one. removed as there shouldn't be concurrent runs as far as I can see possible..

dirkmueller: its gone now.

build-recipe-preinstallimage
```diff
@@ -84,3 +82,23 @@ recipe_cleanup_preinstallimage() {
...
+
+preinstallimage_compress() {
+    local otherdir="$1"
+    local pimage="$otherdir/preinstallimage.tar"
+
+    if test ! -d "$otherdir" -o ! -f "$pimage" -o -L "$pimage" ; then
+        return
+    fi
+
+    case "$(readlink -e -- '$pimage')" in "$(readlink -e -- '$BUILD_ROOT')"*)
```
adrianschroeter, line 94: This is a protection in chroot only?

Do we really need it as the build is anyway not safe... ?

dirkmueller: I think it also protects a symlink being extracted in the VM case. I was concerned that the build result could manipulate something on the build worker host by following symlinks.

mlschroe (approved):

PR #778, opened 2022-02-03
Use faster VM shutdown alternative

dirkmueller:
In most cases a remount -o ro / succeeds wwhich should force flushing
of writes. if it fails we can wait for dirty to be zero.

Typically this reduces the sleep 3 in that code path to 0.1-0.2s.


adrianschroeter:
the sleep was there to ensure that the terminal output got flushed, this seems not to be the case anymore?

dirkmueller:
I see no issues with missing output with virtio-serial. want me to put a sleep in there if its not virtio?

dirkmueller:
@adrianschroeter added logic to check for the logging tee to finish and optionally kill it.

PR #791, opened 2022-02-14
Fix "Can't locate PBuild/Download.pm"

dcermak:
Download.pm has been moved to Build via 55450de8c4b219bdf63358ee2840cfad55409838


adrianschroeter:
you have a point, but there is also a PBuild::Download::fetch reference below, can you complete it?

dcermak:
Good catch @adrianschroeter, I totally forgot to check for the actual usage… It should be corrected now.

PR #793, opened 2022-02-17
Fix vc requiring which

dcermak:


dcermak:
Friendly bump :-)

Conan-Kudo (approved):

adrianschroeter:
had to revert the dependencies patch as it broke installation on most distros.

Shouldn't be a requires anyway since the build script is not needing, maybe with Recommends.

PR #797, opened 2022-02-23
Update vc - default to 80 char width for .changes

zaitor:
Most terminals on openSUSE today defaults to 80 char width, hence set separator line for .changes entries to the same.


coolo:
I don't think we should change that. I bet there are quite some tools out there that parse the changes files as is

adrianschroeter:
it would need a policy discussion for the code streams first at least..., closing it for now. Please reopen when the release managers have agreed to the change.

PR #807, opened 2022-03-23
Neither /sbin/ip nor /sbin/ifconfig might be installed

adrianschroeter:
We still might to have lo network if initrd took take care, like
it does on our current distros


coolo:
We have tests in build? wow :)

PR #812, opened 2022-04-16
Preinstallimage build uid fix

seife:
Do not package abuild user in preinstallimage, so that build-uid=calller can work.
Fix it up in build script so that old preinstallimages also work.

### DIFF
--- a/build
+++ b/build
@@ -1686,8 +1686,10 @@ for RECIPEPATH in "${RECIPEFILES[@]}" ; do
 		egrep "^abuild:" <$BUILD_ROOT/etc/passwd
 		echo "build script attempting to use:"
 		echo "abuild::${ABUILD_UID}:${ABUILD_GID}:..."
-		echo "build aborting"
-		cleanup_and_exit 1
+		echo "fixing..."
+		sed -i -e "/^abuild:/d;iabuild:x:${ABUILD_UID}:${ABUILD_GID}:Autobuild:/home/abuild:/bin/bash" $BUILD_ROOT/etc/passwd
+		sed -i -e "/^abuild:/d;iabuild:x:${ABUILD_GID}:" $BUILD_ROOT/etc/group
+		chown "$ABUILD_UID:$ABUILD_GID" $BUILD_ROOT/home/abuild
 	    fi
 	fi
 	if test -f $BUILD_ROOT/etc/shadow ; then
--- a/build-recipe-preinstallimage
+++ b/build-recipe-preinstallimage
@@ -57,6 +57,11 @@ recipe_build_preinstallimage() {
       esac
       TOPDIRS="$TOPDIRS $DIR"
     done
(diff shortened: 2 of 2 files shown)


adrianschroeter:
our preinstallimages never have abuild user inside, they are built as root. So this seems not to apply to our code base.

seife:
Wrong.
Please do:
```
osc co home:seife:testing collectd
cd home:seife:testing collectd
osc build --clean 15.4
```

Result:
```
[   33s] strace-5.14-150400.1.7                ########################################
[   33s] rzsz-0.12.21~rc-150000.3.3.2          ########################################
[   33s] obs-service-recompress-0.3.1+git201707########################################
[   33s] libnetlink-devel-5.14-150400.1.8      ########################################
[   33s] libmnl-devel-1.0.4-1.25               ########################################
[   33s] less-590-150400.3.3.1                 ########################################
[   33s] regenerate-initrd-posttrans: dracut is not installed, not rebuilding the initrd
[   34s] regenerate-initrd-posttrans: dracut is not installed, not rebuilding the initrd
[   34s] now finalizing build dir...
[   34s] ... running 01-add_abuild_user_to_trusted_group
[   34s] ... running 02-set_timezone_to_utc
[   34s] ... running 11-hack_uname_version_to_kernel_version
[   34s] abuild user present in the buildroot (/local/abuild/osc/15.4-x86_64) but uid:gid does not match
[   34s] buildroot currently using:
[   34s] abuild:x:399:399:Autobuild:/home/abuild:/bin/bash
[   34s] build script attempting to use:
[   34s] abuild::10329:100:...
[   34s] build aborting
[   34s] 
[   34s] strolchi failed "build collectd.spec" at Tue Apr 25 13:08:01 UTC 2023.
[   34s] 
```

Preinstallimage is in home:seife:testing/preinstallimage-base
```
seife@strolchi:~/buildservice/testing/preinstallimage-base> cat _preinstallimage 
BuildRequires: bash
#!BuildIgnore: brp-trim-desktopfiles
#!BuildIgnore: cputype-armv6
```

please reopen.

adrianschroeter:
okay, it is indeed there, but it gets added by a hook from post-build-checks package.

The build script can not overwrite that though. The real fix must be applied in that package then.

seife:
So I could work around it by excluding post-build-checks from preinstallimage.
Anyway, this is damn fragile if any package can beak the preinstallimage.
And this is why there are two commits: the one for build-recipe-preinstallimage which just makes sure that the preinstallimage is not containing the abuild user (which works perfectly fine on its own if this version is installed on the OBS server) and the second one which fixes this up after the fact in the build script used by "osc build" for example. At least the first commit https://github.com/openSUSE/obs-build/pull/812/commits/ff13a5254c38c8c21faecef8a9a99797e0877b2b should be still included to make the preinstallimage not break on users for stupid reasons.

PR #813, opened 2022-04-18
deb: defer dpkg triggers until all packages are installed, and disable man-db altogether

bluca:
man-db takes a lot of time when a lot of packages are installed, and it's really pointless for package builds.
dpkg triggers are intended to be ran after a batch of packages have been installed, and that's how `apt` and friends run them, but here `dpkg --install` is ran manually so the triggers run after each individual package install, which is slow. Run dpkg with `--no-triggers` if it's available. The existing `dpkg --configure` call at the end of the installation will run the triggers.

### DIFF
--- a/build-pkg-deb
+++ b/build-pkg-deb
@@ -61,6 +61,17 @@ deb_setup() {
 exit 101
 EOF
     chmod 755 $BUILD_ROOT/usr/sbin/policy-rc.d
+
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

bluca:
@adrianschroeter ping

PR #818, opened 2022-05-19
Unify debian release patching in debtransform

darix:
We also handle the OBS-DCH-RELEASE tag in debtransform now.


bluca:
Uhm but debtransform is not always ran, no?

darix:
it should be run all the time. the code is run unconditional right after the sed that i removed.

darix:
I actually will try to move the changes file patching into debtransform. and maybe add an option to implement this as part of debtransform:

https://nordisch.org/posts/the-debian-changelog-and-obs/

PR #823, opened 2022-05-28
configs/*.conf: Stop using rpmlint-mini and rpmlint-Factory.

sunweaver:
This relates to https://bugzilla.opensuse.org/show_bug.cgi?id=1187749

 It resolves failing dependency resolving during build env setup:

 This is also reproducible on an openSUSE 15.4 system:

 ```
 Loading repository data...
 Reading installed packages...
 Resolving package dependencies...

 Problem: nothing provides 'rpmlint-mini' needed by the to be installed rpmlint-Factory-1.0-3.19.noarch
  Solution 1: do not install rpmlint-Factory-1.0-3.19.noarch
  Solution 2: break rpmlint-Factory-1.0-3.19.noarch by ignoring some of its dependencies

 Choose from above solutions by number or cancel [1/2/c/d/?] (c): Resolving dependencies...
 Resolving package dependencies...
 Nothing to do.
 ```

### DIFF
--- a/configs/sl11.1.conf
+++ b/configs/sl11.1.conf
@@ -27,7 +27,7 @@ Support: texinfo timezone util-linux login
 Support: libgomp43 libuuid1 psmisc
 Support: terminfo-base
 
-Support: build brp-check-suse post-build-checks rpmlint-Factory
+Support: build brp-check-suse post-build-checks rpmlint
 Keep: brp-check-suse
 
 %ifarch ia64
--- a/configs/sl11.2.conf
+++ b/configs/sl11.2.conf
@@ -45,7 +45,7 @@ Support: terminfo-base
 
 # build tools packages
 Support: build brp-check-suse post-build-checks
-Support: rpmlint-Factory
+Support: rpmlint
 Keep: brp-check-suse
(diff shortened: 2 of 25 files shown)


lnussel:
If rpmlint-mini is not in Leap 15.4 repos then that's a bug in the Leap publish settings. The configs come from OBS so are mostly what is used to build the distro on server side. Adding local modifications should be avoided. You can't just replace rpmlint-mini with rpmlint anyway as `rpmlint-mini` is basically a special plugin for the build script whereas `rpmlint` is a regular package with no relation to the build script. So it wouldn't get called after build.
So I think this request needs to be declined. Please open a bug for Leap to publish rpmlint-mini instead.

adrianschroeter:
while Ludwig is right, I take this change now anyway until the distro got fixed.

lnussel:
:facepalm: as I said that turns rpmlint in the build root to a noop. Ie turning off the extra QA step.  If that is the goal then just remove rpmlint-mini without adding rpmlint. Same effect, except that python isn't pulled into every single build root, spoiling the results.
Was a bug filed at all?

Please revert.

Issue #826, opened 2022-06-03
[RFC] Support address sanitizers within VMs via memory overcommit


Ionic:
I'm creating this issue as a discussion platform. A linked PR will follow up soon.

### Original issue

Currently, packages utilizing asan fail to build within - whether built in a VM or directly on the host. Thus, most packages employ hacks to disable the usage of asan, as done for [meson's test suite](https://build.opensuse.org/package/view_file/openSUSE:Factory/meson/meson.spec?expand=1) (lines 165+ currently, can't link to specific line numbers on OBS).

### Reason

This is because asan tries to opportunistically allocate multiple TB of memory, even when used on a simple hello-world program. That sounds rather weird at first, but the Linux kernel actually allows such behavior heuristically (though it can be tuned via kernel parameters), since it doesn't actually allocate the memory a process requests up front. Instead, it just assigns the region to the process and handles memory allocation on-demand via page faults when first accessed. Most programs requesting huge regions, including asan, never actually use even a medium portion of this region, so just allowing overcommitment is fine.

Now, that only works if `RLIMIT_AS` is higher or equal to the requested memory size. if it's set to `infinity`, that's trivially true, however, [build actually limits it to `(2/3) * (total_mem + total_swap)` in stage1](https://github.com/openSUSE/obs-build/blob/376b661351cd62511fe68ff225f09a3c754a7c58/build#L743), which makes sense, since `stage1` runs natively on the host (and does things like setting up the VM if necessary, preinstalling packages and the like).

Then, in `stage2`, which is executed within the VM, [build skips setting the memory limit](https://github.com/openSUSE/obs-build/blob/376b661351cd62511fe68ff225f09a3c754a7c58/build#L726), but at that point, it's too late, since the parent process that spawned the VM (`stage1`) already had a limit set which is passed down to the VM process.

This issue might not actually affect all VM types, as far as I've seen, since types such as `ec2` or `zvm` are spawned remotely and not influenced by any `PRLIMIT` set on the builder host machine.

### Remedy

We should be able to make that work by removing the limit imposed on VM processes either via the VM's spawn tools (if they support modifying prlimits natively) or by setting it to `unlimited` in shell right before spawning the VM. This is essentially what my PR is doing.

### Impact

That is a bit difficult to tell. I assume that the memory limit was imposed to protect the host system from malicious (or just buggy) code executed within the builders, i. e., so as to not make a builder unresponsive or trigger the OOM watchdog.

[...]

PR #827, opened 2022-06-03
[VM] Override address size limit to unlimited

Ionic:
The "build" script limits PRLIMIT_AS to (2/3) * (mem_size + swap_size),
which breaks asan, which tries to opportunistically allocate multiple TB
of memory, but crucially never even closely uses this.

This commit eliminates the memory limit right before spawning VMs (and
containers) if necessary, so that stage2 can run unrestricted
(notwithstanding the memory limit imposed by hypervisors) and use memory
overcommitment within the build process.

See the linked bug for a discussion of this.

Fixes: https://github.com/openSUSE/obs-build/issues/826

### DIFF
--- a/build-vm-docker
+++ b/build-vm-docker
@@ -32,6 +32,9 @@ vm_startup_docker() {
     docker rm "$name" >/dev/null 2>&1 || true
     local docker_opts=
     test -n "$VM_TYPE_PRIVILEGED" && docker_opts="--privileged --cap-add=SYS_ADMIN --cap-add=MKNOD"
+    # Override memory restriction set by stage1. Docker does not support
+    # --ulimit as=... for some reason.
+    ulimit -v 'unlimited'
     docker run \
         --rm --name "$name" --net=none $docker_opts \
         --mount "type=bind,source=$BUILD_ROOT,destination=/mnt" \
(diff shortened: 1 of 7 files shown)


adrianschroeter:
the ulimit calls would fail when using build script as non-root?

I agree that a limit when doing VM builds seem not to be necessary, not sure about this approach though.

dirkmueller:
is this still needed after https://github.com/openSUSE/obs-build/commit/52cec996150a0050ad0c873cc4fc750a35fd460c ?

Ionic:
> the ulimit calls would fail when using build script as non-root?

Probably, yes, but doesn't spawning a build require root access to begin with? The actual build process is being run unprivileged in most cases, but the setup typically requires elevated privileges.

> is this still needed after https://github.com/openSUSE/obs-build/commit/52cec996150a0050ad0c873cc4fc750a35fd460c ?

Very good question. That one (obviously) is completely new to me. Looking at it and reading up, it does look like my PR is obsolete after the mentioned commit and that it fixes asan usage within VMs/containers. I'll test that and close the PR and bug report if it turns out to essentially allow memory overcommitment.

Ionic:
Unfortunately, even https://github.com/openSUSE/obs-build/commit/52cec996150a0050ad0c873cc4fc750a35fd460c breaks address sanitizers with LXC builds.

After that change, I'll have to override `lxc.prlimit.data` to get packages with address sanitizers working, of course, but it essentially boils down to the same breakage.

I can't really test the other virtualization/containerization options (I'd prefer using Qemu/KVM in the first place, but the VM I'm running an OBS appliance on doesn't support nested virtualization, so I'm out of luck on that front).

Will update this PR accordingly.

Issue #830, opened 2022-06-10
Regression in debian obs_scm builds because of b2b0430


darix:
Since the deployment of tag 20220531 the [darktable master snapshot][1] builds are failing.

I started bisecting this issue going from tag 20220429. Afterwards i tested with HEAD from master and [b2b0430][2] by @ila-embsys  reverted. that fixes the build as well.

[1]: https://build.opensuse.org/package/show/graphics:darktable:master/darktable
[2]: https://github.com/openSUSE/obs-build/pull/760

ila-embsys:
Thanks for the feedback!

The 'debtransform' tool had ignored the 'Format' tag in *.dsc file and had always interpreted it as 'Version 1.0' before the b2b0430 merged. Try to change into *.dsc the value of the 'Format' tag from '3.0 (quilt)' to '1.0' to get previous behaviour.

Here is [the fixed build](https://build.opensuse.org/package/show/home:ila.embsys:branches:graphics:darktable:master/darktable).

darix:
well the point is more that we would need to announce that breaking change. as the current code is not drop in.

or even better, fix the code that it wouldnt break existing packages.

ila-embsys:
I think I have understood where the problem is.

In the building's log 'darktable' package, the actual version is '3.9.0' but the build failed with an error because of not found version '2.4.0'.

The only place where I found this version referred to is the last entry of the 'changelog'.

I don't know why the builder is trying to use this version for the file name. However, previously it could work correctly because during the creation of *.diff file for DSC 1.0, the 'debtransform' was adding an entry to the 'changelog' with the actual version.

I didn't notice such behaviour and didn't provide for DSC 3.0 the same handling in 'changelog' entries.

darix:
without your patch the version number gets patched and it finds the proper tarball. with your patch it is not patched so it looks for an outdated tarball.

darix:
maybe this line no longer gets executed: https://github.com/ila-embsys/obs-build/blob/eff348e15368b6b8be80d55f33d673edcd5f2770/build-recipe-dsc#L138

 (i mean the whole patching of changelog file should be in debtransform anyway)

munix9:
I have the same problem since recently:
https://build.opensuse.org/project/monitor/home:munix9:unstable

It searches for version 1.7.0\~git20190807+d9746691 although the current version is 1.7.0~git20220611+3ccb02b.

No idea how the service gets to "1.7.0~git20190807+d9746691".

[...]

PR #831, opened 2022-06-12
Add Changelog patching when building DSC format 3.0.

ila-embsys:
'debtransform' is patching 'changelog' when make *.diff for DSC 1.0. It adds a 'changelog' entry with the actual build package version.

Missing the actual build package version in 'changelog' is the reason for misbehavior because after 'debtransform' finishes its work, the version from the last 'changelog' entry is used for determining the package version.

This merge request contains changes to maintain 'changelog' patching during the build DSC 3.0 in the same way as for DSC 1.0.
To do it, the patching procedure was separated into a separate function `dopatch_changelog()` then used for both building scenarios.

In case of building DSC 3.0 added 'changelog' catching in two different cases:
- When adding a file with the name "debian.changelog" `dofile_quilt()`.
- When adding tar contents containing "debian/changelog" `dotar_quilt()`.

Also added test number 5. It contains the 'changelog' with the outdated version in the last entry in 'debian.tar.gz'. Test checks that after 'debtransform' finishes its work, the 'changelog' is patched and comes with a new entry with the actual build package version. It is the case of a broken building of packages, as mentioned in #830.


darix:
+1 confirmed with darktable testbuild

Issue #833, opened 2022-06-15
debtransform fails in step "add tar file contents"


repagh:
I get these kind of errors:

[  150s] Add tar file contents 'debian.tar'
[  150s]     to debian quilt tar file '../SOURCES.DEB/linux-preempt_5.10.73rt54-1+3.1.debian.tar'
[  150s] tar: /tmp/Debtej7BQC/debian.tar: Cannot open: No such file or directory

on a previously-working project. There is a debian.tar file in the project, and a local build (from Fedora 34, with probably an older debtransform) simply works. On the older script, the debian.tar is copied to the SOURCES folder, and then merged with the source tarball.

project is here:

https://build.opensuse.org/package/live_build_log/home:repabuild:preempt/preempt_rt5_10/xUbuntu_20.04/x86_64

PR #840, opened 2022-07-03
Add handling of non-compressed tar when creating Debian archive for DSC 3.0

ila-embsys:
Before these changes, the function `dotar_quilt()` that creates the Debian archive unpacked compressed archives like `gz`, `xz`, `bz2` to a `temp` folder and used those uncompressed `tar`s for concatenation to the final archive. Then the function was removing the intermediate file at the end.

This patch adds possibility to pass to the `dotar_quilt()` function an uncompressed `tar` archive. The function skips the copying of the uncompressed `tar` to the `temp` folder and uses the original path for concatenation until a `changelog` in the uncompressed  `tar` found.

Also added a test number 6 for this use case.

The patch should fix https://github.com/openSUSE/obs-build/issues/833 and fix https://github.com/openSUSE/obs-build/issues/842


debtransform
```diff
@@ -270,6 +278,13 @@ sub dotar_quilt {
     unlink($changelog_file_path);
     rename($changelog_file_path_patched, $changelog_file_path);
 
+    # Copy to temp dir if not already copied
+    if ((index ${tar_for_concatination_path}, ${tempdirname}) == -1){
+      print "Copy '${tar_for_concatination_path}' to '${tempdirname}' for patching\n";
+      system("cp \"$decompress_src\" \"$decompress_dst\"") == 0 || die("Can not copy archive '$decompress_src': $!\n");
```
darix, line 284: Call it with an list as param.

debtransform
```diff
@@ -279,7 +294,11 @@ sub dotar_quilt {
   }
 
   system("tar --concatenate -f \"$output_file_path\" \"${tar_for_concatination_path}\"");
-  unlink("${tar_for_concatination_path}");
+
+  # Check that tar for concatination in temp dir then remove it
+  if ((index ${tar_for_concatination_path}, ${tempdirname}) == 0){
+    unlink("${tar_for_concatination_path}");
```
darix, line 300: why the quotes around the variable in the unlink() call?

debtransform
```diff
@@ -233,17 +233,25 @@ sub dotar_quilt {
...
-      system("gzip -dck \"$decompress_src\" > \"$decompress_dst\"") == 0 || die("Can not unpack gzip archive '$decompress_src': $!\n");
-  } elsif (($ext eq ".tar.xz")) {
-      system("xz -dck \"$decompress_src\" > \"$decompress_dst\"") == 0 || die("Can not unpack xz archive '$decompress_src': $!\n");
-  } elsif (($ext eq ".tar.bz2")) {
-      system("bzip2 -dck \"$decompress_src\" > \"$decompress_dst\"") == 0 || die("Can not unpack bzip2 archive '$decompress_src': $!\n");
-  }
+  my $tar_for_concatination_path;
+
+  if (($ext eq ".tar")) {
+      $tar_for_concatination_path = $input_file_path;
```
mlschroe, line 239: I wonder if it wouldn't be easier to just do a 
```
  system('cp', $decompress_src, $decompress_dst);
```
here to keep the code simple.

debtransform
ila-embsys: In all 'system' function calls, the strings have been replaced with an argument list. In places where pipes were, 'system' function calls have been replaced with IPC::Run.

debtransform
ila-embsys: Removed it.

debtransform
ila-embsys: Good idea. It simplifies code reading.

bluca:
FYI @ila-embsys @adrianschroeter this PR is breaking all deb-based builds that use debtransform:

```
[  156s] Can't locate IPC/Run.pm in @INC (you may need to install the IPC::Run module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.28.1 /usr/local/share/perl/5.28.1 /usr/lib/x86_64-linux-gnu/perl5/5.28 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl/5.28 /usr/share/perl/5.28 /usr/local/lib/site_perl /usr/lib/x86_64-linux-gnu/perl-base) at /.build/debtransform line 216.
[  156s] BEGIN failed--compilation aborted at /.build/debtransform line 216.
[  156s] debian transforming failed.
[  156s] ### VM INTERACTION START ###
```

The IPC module is not part of perl-base, but it's in a separate package, `libipc-run-perl` so all build definitions need to be updated to require it.

Eg: https://build.opensuse.org/package/live_build_log/home:bluca:dpdk/dpdk-21.11/Debian_10/x86_64

adrianschroeter:
right, needed to revert that one. sorry.

darix:
@adrianschroeter why not just add the missing requires to the debian package?

adrianschroeter:
that has no effect during build.

all prjconf would need to get altered, but that is not a compatible change.

bluca:
it seems it's just to run the decompressors, but it was using `system()` before just fine? So maybe just resubmit a new revision that keeps `system()`?

PR #857, opened 2022-08-15
Add Docker::query && Docker::queryhdrmd5

dcermak:
This is a draft/WIP to add support to write .bininfo files for docker based builds.

Currently this is very hacky and works only to a certain extend: `pbuild` writes the `.bininfo` files, but it is still not able to resolve dependencies from that. I have tested this against https://github.com/dcermak/bci-sle-15-sp4 where all containers except micro and base are unresolvable because they depend on base, but somehow nothing provides `container:suse/sle15:15.4`…

### DIFF
--- a/Build/Docker.pm
+++ b/Build/Docker.pm
@@ -20,10 +20,14 @@
 
 package Build::Docker;
 
+
 use Build::SimpleXML;	# to parse the annotation
 use Build::SimpleJSON;
 
 use strict;
+
+eval { require JSON; };
+*JSON::decode_json = sub {die("JSON::decode_json is not available\n")} unless defined &JSON::decode_json;
 
 sub unify {
   my %h = map {$_ => 1} @_;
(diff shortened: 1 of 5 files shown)


dcermak:
I have found the issue, I did not update `binsufs` in `LocalRepo.pm` and also did not include the `source` key in the `.bininfo` hash

dcermak:
Well, turns out this is not everything. pbuild manages to schedule builds, but Dockerfile based builds fail with:
```
[  106s] -----------------------------------------------------------------
[  106s] ----- building Dockerfile
[  106s] -----------------------------------------------------------------
[  106s] -----------------------------------------------------------------
[  106s] find: 'containers': No such file or directory
[  106s] base image not found
```

dcermak:
I got a bit further, I can now rebuild containers based on kiwi images, but for some reason dockerfile based containers are not included in the dependency resolution yet…

Build/Docker.pm
```diff
@@ -20,6 +20,8 @@
 
 package Build::Docker;
 
+use JSON;
```
adrianschroeter, line 23: why is SimpleJSON not enough?

If really needed, check how the YAML parser in AppImage parser is loaded optional...

dcermak: SimpleJSON cannot parse json, it can only convert perl data structures to JSON, but I need to convert the .containerinfo files into perl hashes.

I have adjusted the code accordingly.

dcermak:
I have got it to work, pbuild now appears to correctly schedule container builds that depend on each other.

Build/Docker.pm
```diff
@@ -78,6 +82,98 @@ sub quote {
...
+    # - for kiwi builds we're lucky as the containerinfo contains the property
+    #   'name' and just about everything else that we need
+    # - with dockerfiles we're a bit screwed, as the package name is "lost" in
+    #   the build process and the archive filename is picked from the first tag...
+    #   So we just pick that as the name and hope things don't break
+    my $name = $data->{'name'};
+    my $version = $data->{'version'};
+    my $release = $data->{'release'};
+
+    my $name = $data->{'name'};
```
Vogtinator, line 113: Duplicate lines?

dcermak: yes :facepalm:

bluca:
@dcermak nice stuff!

Given you managed to make this work, I was wondering if you could please give me a hint - I'm adding a new image format here: https://github.com/openSUSE/obs-build/pull/861

That works, so I also tried to add image dependencies (the tool supports building images based on other images), but I can't convince the resolver to resolve the dependencies, and I can't figure out what I'm missing. I've implemented query/queryhrdmd5 and added the extensions in a few places: https://github.com/bluca/obs-build/commit/0ef7992e70f512148fb6c
The metadata looks good when inspecting the built images through the web UI, I see the name, the provides and so on. But the new image build with the dependency remains unresolvable.

Do you have any suggestion? Thanks!

dcermak:
> The metadata looks good when inspecting the built images through the web UI, I see the name, the provides and so on. But the new image build with the dependency remains unresolvable. With some debug Data::Dumper I can see the returned hash from query() looks good.
> 
> Do you have any suggestion? Thanks!

I'm everything **but** a pbuild expert, so the only thing that I can recommend is to add debug statements all over https://github.com/openSUSE/obs-build/blob/653c29d329d5172202c438d6b54a70538e2ba8f3/PBuild/LocalRepo.pm#L149
and see if this function gets called at all and whether it aborts/skips your new image format somewhere.

I put a `print Dumper \$var;` more or less in every important line and would fire off builds until I found out why the heck things weren't working…

bluca:
> > The metadata looks good when inspecting the built images through the web UI, I see the name, the provides and so on. But the new image build with the dependency remains unresolvable. With some debug Data::Dumper I can see the returned hash from query() looks good.
> > Do you have any suggestion? Thanks!
> 
> I'm everything **but** a pbuild expert, so the only thing that I can recommend is to add debug statements all over
> 
> https://github.com/openSUSE/obs-build/blob/653c29d329d5172202c438d6b54a70538e2ba8f3/PBuild/LocalRepo.pm#L149
> 
> 
> and see if this function gets called at all and whether it aborts/skips your new image format somewhere.
> I put a `print Dumper \$var;` more or less in every important line and would fire off builds until I found out why the heck things weren't working…

Thanks, that's very useful, will do that. What was the root cause in the end for your case?

dcermak:
> Thanks, that's very useful, will do that. What was the root cause in the end for your case?

A bit of everything, I was missing the `sources` field, the name extraction for Dockerfile based builds was totally busted and the binsufs array did not contain the correct file endings for containers, making pbuild simply skip them -.-

bluca:
I'm getting zero hits on a bunch of debug dumps sprinkled over LocalRepo.pm - does the Repotype have to be set to any specific value in prjconf? I really need to figure out how to run pbuild under the Perl debugger...

mlschroe:
It shouldn't be necessary to add a querthdrmd5 function, the data is supposed to come from the containerinfo file.

mlschroe:
(Or, like in OBS, via a .obsbinlnk file.)

dcermak:
> It shouldn't be necessary to add a querthdrmd5 function, the data is supposed to come from the containerinfo file.

IIRC it is necessary to have a `queryhdrmd5`, as otherwise the following call in `read_bininfo` fails:
https://github.com/openSUSE/obs-build/blob/653c29d329d5172202c438d6b54a70538e2ba8f3/PBuild/BuildResult.pm#L84

mlschroe:
Now implemented in 50b1a5881130d2d26ca907874f4f9f37eeea8a06

PR #859, opened 2022-08-16
Add RemoteAsset support for Dockerfile based builds

dcermak:


docs/source_assets.adoc
```diff
@@ -50,7 +50,16 @@ disabled via
...
 ==== Dockerfile description
 
-Has no support for assets yet.
+Remote assets that will be downloaded before the build can be added via the
+comment `#!RemoteAssetUrl: https://my.url/asset.tar`. `pbuild` will fetch the
+asset before the build and will make it available in the buildroot so that the
+asset can copied into the container as follows:
+[source,Dockerfile]
+----
+#!RemoteAssetUrl: https://my.url/asset.tar
```
adrianschroeter, line 59: can you add an example how to do it using a sha256 checksum here as well please?

dcermak:
Adrian Schröter ***@***.***> writes:

> @adrianschroeter commented on this pull request.
>
>
>
>> @@ -50,7 +50,16 @@ disabled via
>  
>  ==== Dockerfile description
>  
> -Has no support for assets yet.
> +Remote assets that will be downloaded before the build can be added via the
> +comment `#!RemoteAssetUrl: https://my.url/asset.tar`. `pbuild` will fetch the
> +asset before the build and will make it available in the buildroot so that the
> +asset can copied into the container as follows:
> +[source,Dockerfile]
> +----
> +#!RemoteAssetUrl: https://my.url/asset.tar
>
> can you add an example how to do it using a sha256 checksum here as well please?

That part is not implemented by this PR. I tried to, but it didn't
really work, so I removed it again.

dcermak:
@adrianschroeter Can we move this forward?

PR #861, opened 2022-08-21
Add new image format: mkosi

bluca:
Modern image builder from the systemd project:
https://github.com/systemd/mkosi

Supports rpm, deb, arch (and gentoo) based distros. Capable of building
an image suitable to be booted via UEFI on baremetal, virtual machine
and container, using various filesystems including dm-verity, following
the Discoverable Partitions Specification:
https://systemd.io/DISCOVERABLE_PARTITIONS/
Also can build system extensions:
https://www.freedesktop.org/software/systemd/man/systemd-sysext.html


lnussel:
awesome! :beers:

mlschroe:
Thanks!

bluca:
@mlschroe thanks for the merge! The backend side is here: https://github.com/openSUSE/open-build-service/pull/12972

PR #872, opened 2022-09-07
mkosi: implement obsbinlnk

bluca:
Allow to build mkosi images using other mkosi images (systemd-sysext)


bluca:
@mlschroe @adrianschroeter need some help as this is not yet working, and I can't figure out what the problem is - I followed the same pattern as https://github.com/openSUSE/obs-build/commit/50b1a5881130d2d26ca907874f4f9f37eeea8a06 but still I can't get an .obsbinlnk file. I also tried to implement query() but that also doesn't help.

What am I missing?

adrianschroeter:
Can you give us a setup (either via git or somewhere in build.o.o) where we could have a look? basically a two mkosi image build setup.

bluca:
@adrianschroeter sure, this is the base:

https://build.opensuse.org/package/show/home:bluca:mkosi/suse-image

and this is the extension that should pull the base in (`BaseImage=base`):

https://build.opensuse.org/package/show/home:bluca:mkosi/suse-sysext

With the patches from this PR applied, the build stays unresolvable in my local VM.

I've also pushed a patch to add `queryiso()` as the doc suggests, but still no luck.

Thank you!

bluca:
@adrianschroeter did you have a chance to have a look at this?

bluca:
@adrianschroeter @mlschroe gentle ping on this issue

lnussel:
looks like this is almost working in pbuild. The code produces a .obsbinlnk file with content like this:

```
{
  arch     => "x86-64",
  hdrmd5   => "57c85225a2ed7904531814a64fe99c4a",
  lnk      => "base",
  name     => "base",
  provides => ["opensuse:tumbleweed", "base = ", "suse-image = "],
  release  => 0,
  source   => "suse-image",
  version  => undef,
}
```

I think the name needs to be prefixed with something like `mkosi:`, ie `mkosi:base` to indicate a namespace. Otherwise the dependency resolver cannot know that `base` is not a package.

Build/Mkosi.pm
```diff
@@ -67,11 +67,75 @@ sub parse {
   if (length $cfg->val('Content', 'BuildPackages')) {
     push(@packages, split /\s+/, $cfg->val('Content', 'BuildPackages'));
   }
+  if (length $cfg->val('Partitions', 'BaseImage')) {
+    push(@packages, $cfg->val('Partitions', 'BaseImage'));
```
lnussel, line 71: here it whould add the prefix `mkosi:` I think

bluca:
> looks like this is almost working in pbuild. The code produces a .obsbinlnk file with content like this:
> 
> ```
> {
>   arch     => "x86-64",
>   hdrmd5   => "57c85225a2ed7904531814a64fe99c4a",
>   lnk      => "base",
>   name     => "base",
>   provides => ["opensuse:tumbleweed", "base = ", "suse-image = "],
>   release  => 0,
>   source   => "suse-image",
>   version  => undef,
> }
> ```
> 
> I think the name needs to be prefixed with something like `mkosi:`, ie `mkosi:base` to indicate a namespace. Otherwise the dependency resolver cannot know that `base` is not a package.

I've added that - do you know where the side that consumes it lives, to make it parse that?

lnussel:
Got it working with something that looks like this:
```
{
  arch     => "x86-64",
  hdrmd5   => "f1fa8d13dfbbc294855ee8bbf7b01261",
  lnk      => "base.raw",
  name     => "mkosi:base",
  provides => [
                "opensuse:tumbleweed",
                "mkosi:base = 0",
                "mkosi:suse-image = 0",
              ],
  release  => 0,
  source   => "suse-image",
  version  => 0,
}
```

Consumer would be `PBuild::RepoMgr::copyimagebinaries` after adding a code path for mkosi :) See pr#985
However, this is only for pbuild. Very similar code is in OBS. PBuild is a stripped down clone basically.

PR #875, opened 2022-09-16
[build-vm] use poweroff instead of halt -p

dcermak:
Some minimal coreutils installations like toybox provide halt, but it does not support the -p option. As halt -p is equivalent to poweroff, we use that instead and can then build toybox successfully


adrianschroeter:
okay, poweroff seems to exist long enough, at least since SLE 10

PR #876, opened 2022-09-18
build-recipe-livebuild: run as root with 'sh -c' instead of 'su -c ... - root'

bluca:
livebuild stopped working some time ago:

[   48s] -----------------------------------------------------------------
[   48s] ----- building standard.livebuild (user abuild)
[   48s] -----------------------------------------------------------------
[   48s] -----------------------------------------------------------------
[   49s] su: Critical error - immediate abort
[   50s] su: Critical error - immediate abort

https://build.opensuse.org/package/live_build_log/home:bluca:debian_secure_boot/example-images/img_nokernel/x86_64

Run it as root instead of abuild like other image build systems, and use sh -c.

### DIFF
--- a/build-recipe-livebuild
+++ b/build-recipe-livebuild
@@ -70,7 +70,8 @@ createrepo_debian() {
 	gzip -c9 dists/${DIST}/main/source/Sources \
 	    > dists/${DIST}/main/source/Sources.gz
 	EOF
-    chroot $BUILD_ROOT su -c "sh /.createrepo_debian.tmp.sh" - root
+    chmod +x ${BUILD_ROOT}/.createrepo_debian.tmp.sh
+    chroot $BUILD_ROOT /.createrepo_debian.tmp.sh
     local RESULT=$?
     rm -f $BUILD_ROOT/.createrepo_debian.tmp.sh
     [ "${RESULT}" != 0 ] && return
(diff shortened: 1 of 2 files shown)


build-recipe-livebuild
```diff
@@ -133,7 +133,7 @@ recipe_build_livebuild() {
     # the example cited in the above bug, the metadata for sid would be incorrect
     # anyway, and we would want the ones for potato.
     if test "${DIST}" = "n/a" ; then
-	DIST=$(chroot $BUILD_ROOT su -c "sed 's/\(.*\)\/.*/\1/' /etc/debian_version")
+	DIST=$(chroot $BUILD_ROOT sh -c "sed 's/\(.*\)\/.*/\1/' /etc/debian_version")
```
adrianschroeter, line 136: can't you simply drop the   sh -c   part?

build-recipe-livebuild
```diff
@@ -193,9 +193,9 @@ recipe_build_livebuild() {
 
     if [ -x $BUILD_ROOT/usr/lib/build/livebuild_pre_run ] ; then
 	echo "Running OBS build livebuild_pre_run hook"
-	chroot $BUILD_ROOT su -c \
+	chroot $BUILD_ROOT sh -c \
         "export RECIPEFILE=${RECIPEFILE}; /usr/lib/build/livebuild_pre_run" \
```
adrianschroeter, line 197: same here

adrianschroeter: and in the next lines;)

build-recipe-livebuild
bluca: Removed here and in a few other places

build-recipe-livebuild
bluca: Here and below it's running a shell line, not a single command (eg, export or change directory, then run thing), so I don't think I can for this one

adrianschroeter: right

PR #877, opened 2022-09-20
initial support for Google Cloud builds

adrianschroeter:
Not yet ready to merge:
- detect disk device on attach and adapt CLI for it
- fix hardcoded names (vm-2 and friends)
- cleanup

issues to fix for becoming productive
- look into a way how to use disk encryption
- look into a way to setup bootloader without possible conflicts in build environment
- fix random build failures
- improve serial log behaviour
- looking for speedups .... we need currently 3 minutes on new hardware of gcloud versus 20 seconds on old hardware with kvm
- add support to inject sysrq events ... how?
- find a way to avoid the need of a 10GB storage device even for the smallest build


adrianschroeter:
Not yet ready to merge ... the pull request is just for discussion

michaeldtz:
description of topic: look into a way to setup bootloader without possible conflicts in build environment --> kernel needs to be in the build image. In current use cases the kernel is provisioned from outside.

PR #884, opened 2022-09-27
Call ls using its full name

dcermak:
`ls` from `PBuild::Util` is not in the global namespace and referencing it without the full namespace results in errors like:
```
Undefined subroutine &PBuild::Job::ls called at /usr/lib/build/PBuild/Job.pm line 348.
```


mlschroe:
That's weird, didn't that work already at some point in time? Anyway, thanks a lot.

PR #893, opened 2022-10-06
Skip iothreads on QEMU 7.1.0

dirkmueller:
It appears there is a new regression on QEMU 7.1.0 (worked on 7.0) where iothreads together with aio=io_uring hang. As aio=io_uring is slightly faster, we disable iothreads to avoid the hang but still have some performance gain.

### DIFF
--- a/build-vm-kvm
+++ b/build-vm-kvm
@@ -406,8 +406,12 @@ vm_fixup_kvm() {
 
     # move IO into separate I/O thread for some architectures
     if test "${kvm_device%%-*}" = "virtio" ; then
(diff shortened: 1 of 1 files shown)


dirkmueller:
@adrianschroeter we can get a fixed qemu on opensuse but maybe not on other dists. the issue is known upstream: https://gitlab.com/qemu-project/qemu/-/issues/1193

adrianschroeter:
okay..

ptesarik:
OK, does it mean I have to live with the performance hit until a newer QEMU is released? Is there a way to override this workaround if I run a fixed Tumbleweed qemu 7.1.0?

PR #903, opened 2022-12-05
Update configs/arch.conf

luc14n0:
Revamp the file to preinstall the most basic packages that a minimum Arch Linux system expects, but avoiding the addition of packages that are not really needed due to the hermetic nature of our build system. Based on https://archlinux.org/packages/core/any/base/.

Also preinstall the minimum set of packages makepkg expects in a basic build environment, thus excusing them from being explicitly required in a PKGBUILD script. Based on the base-devel package group (https://archlinux.org/groups/x86_64/base-devel/).

Fixes: https://github.com/openSUSE/open-build-service/issues/13403

### DIFF
--- a/configs/arch.conf
+++ b/configs/arch.conf
@@ -1,25 +1,54 @@
 Repotype: arch
 
 ExpandFlags: preinstallexpand
+
 VMinstall: util-linux
 Preinstall: pacman pacman-mirrorlist sed grep
 
-Required: binutils gcc glibc libtool
+###############################################################################
+# Minimal package set to define a basic Arch Linux installation. Based on
+# https://archlinux.org/packages/core/any/base/, with the exclusion of
+# non-useful packages due to the hermetic nature of our own build system.
 
-Support: acl autoconf automake zlib bzip2 filesystem curl
-Support: libtool ncurses perl gpgme libarchive openssl libssh2
-Support: libassuan libgpg-error attr expat xz pacman pacman-mirrorlist
-Support: fakeroot file sudo patch make net-tools pkgconf inetutils
-Support: bison flex gettext which
+Required: gcc-libs glibc
+
+Support: bash bzip2 coreutils file filesystem findutils gawk gettext gzip tar
+Support: xz
+
+###############################################################################
+# Minimal package set that is expected in a basic Arch Linux build environment
+# using makepkg (https://wiki.archlinux.org/title/makepkg#Usage). Based on
+# https://archlinux.org/groups/x86_64/base-devel/, with the exclusion of
+# non-useful packages due to the hermetic nature of our own build system.
+
+Required: gcc
+
+Support: autoconf automake binutils bison debugedit fakeroot flex groff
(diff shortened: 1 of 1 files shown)


luc14n0:
Rather than using:

```
Preinstall: archlinux-keyring bash bzip2 coreutils file filesystem findutils
Preinstall: gawk gcc-libs gettext glibc grep gzip pacman sed tar util-linux xz
```

a simple `Preinstall: base` would suffice, but that would pull questionably useful packages such as:

```
iproute2, iputils, licenses, pciutils, procps-ng, psmisc, systemd, systemd-sysvcompat
```

And can someone tell me, which configs get precedence, the ones in here or the ones from OBS? Or do the both of them add up?

luc14n0:
Gentle ping @adrianschroeter @mlschroe, please. Arch builds are broken, if the user doesn't workaround the issue, requiring `gzip` either directly in the `PKGBUILD` or in the project configuration.

mlschroe:
Regarding precedence: The OBS version is what is used in OBS. The version from the build package is only used if you want to directly run the "build" or "pbuild" commands from this package. Thus, the version here usually lags behind the one in OBS.

mlschroe:
Oh, and the goal is to keep the "Preinstall" package list as small as possible.

luc14n0:
> The "Preinstall" packages are not the "minimum arch linux system", but the packages needed to get the package manager running. So they are typically a subset of the "minimum system".

Ah, OK. Noted!

> The "minimum system" is specified by the "Support" / "Required" lines. The difference is that "Required" packages will be tracked and result in an automatic rebuild if they change. For example, you might want to put a compiler in the required list, because you want an automatic rebuild if the compiler changes. You might want to put "make" in the support list, because a new make version is usually should not trigger a rebuild of all packages.

Yes, I'm aware of the **Required**/**Support** nuances. But, honestly, I was a bit skeptical about OBS triggering rebuilds for such projects using DoD, or it's better to keep this distinction - organize them appropriately under **Required** and **Support** - in the event someone decides to build e.g GCC in their projects for Arch? Makes sense to me.

> Regarding precedence: The OBS version is what is used in OBS. The version from the build package is only used if you want to directly run the "build" or "pbuild" commands from this package. Thus, the version here usually lags behind the one in OBS.

Yeah, right. Makes perfect sense now. Thanks for all the information.

So, the summary is:

* Keep the **Preinstall** to a minimum set, as much as possible, really;
* And make distinction of **Required** and **Support** packages.

On it.

luc14n0:
Expanding on my comment in [open-build-service#13403](https://github.com/openSUSE/open-build-service/issues/13403#issuecomment-1362090261). Here's the [difference between a meta-package and a package-group](https://wiki.archlinux.org/title/Meta_package_and_package_group#Difference_between_meta_package_and_package_group).

If the idea of creating a meta-package based on the `base-devel` package group, sounds interesting, just let me know.

luc14n0:
I tried, to the best of my ability, to separate the packages between **Required** and **Support** there.

luc14n0:
Ping!

PR #905, opened 2022-12-19
Better filetype detection for temp changes files

jcronenberg:
Right now the tempfile when invoking the `osc vc` command has a naming scheme that results in something like `package.changes.vctmp.slkqwo`. This has the problem that editors then can't as easily detect that it is a `.changes` file. This PR changes the naming scheme so that it would be something like `vctmp.slkqwo.package.changes`

### DIFF
--- a/vc
+++ b/vc
@@ -118,7 +118,7 @@ if [ ! -e "$changelog" ]; then
 	touch $changelog
 fi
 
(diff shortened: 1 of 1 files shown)


mlschroe:
Thanks!

PR #910, opened 2023-01-16
Explain why obs-docker-support failed

darix:
Without this it is hard to understand for the user why this happened.

### DIFF
--- a/obs-docker-support
+++ b/obs-docker-support
@@ -314,7 +314,11 @@ EOF
 }
 
 if test `id -u` != "0"; then
-    echo "obs-docker-support: not executed as root user!"
+    echo <<EOF
+obs-docker-support: not executed as root user!
+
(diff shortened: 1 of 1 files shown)


obs-docker-support
```diff
@@ -314,7 +314,11 @@ EOF
 }
 
 if test `id -u` != "0"; then
-    echo "obs-docker-support: not executed as root user!"
+    echo <<EOF
+obs-docker-support: not executed as root user!
+
+This can happen if your base container (see FROM line) is setting USER already. Than builds in dependent container run as that user/group.
```
adrianschroeter, line 320: This sound not really english to me:  "Than builds in dependent container run as that user/group."

What about "In that case the dependent container commands always gets executed as that user and can not modify the container anymore."

PR #916, opened 2023-02-08
Z vm fixes

azouhr:
This is the changes, that I currently use to operate the z/VM worker on OBS. In order to simplify the debugging and logging of the worker, I also added a function ".log" that allows to specify a certain log level and to print out the affected file and function within that file. 

.log 5 "This is a notice"

The severity levels are used from https://en.wikipedia.org/wiki/Syslog#Severity_level
The default severity level is set to 4, to keep the default output clean.

Feel free to do changes if needed.

### DIFF
--- a/build-vm
+++ b/build-vm
@@ -309,6 +309,9 @@ vm_parse_options() {
 }
 
 vm_set_buildstatus() {
+    .log 5 "VM_SWAP: $VM_SWAP"
+    .log 5 "VM_SWAPDEV: $VM_SWAPDEV"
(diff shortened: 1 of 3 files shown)


adrianschroeter:
This way of logging is conflicting with our current idea of debug logging, we do not want to increase or decrease all kinds of logs, but we want to enable logging for specific areas. We want to enable logging by a specified area (eg. zvm, kvm or all vm handling) and enable debug logging only for these kind of operations.

Could you adapt your zvm worker fixes just to check for BUILD_VERBOSE_VM=true for example?

PR #923, opened 2023-03-01
build: add '--recipe-opt=OPTIONS' to pass generator options

### LINKED ISSUE #922: No way to pass custom options to the generator tool
I tried to debug the rpmbuild dependency generator for mingw builds on openSUSE with osc. 

For this, the `rpmbuild` command has the command line option `--rpmfcdebug`. Unfortunately, there is no way to pass this option to rpmbuild via `osc`. You can pass command line parameters to `build` with `osc --build-opt=xxx`:

```sh 
$ osc build --help | grep opt
    --build-opt=OPT     pass option OPT to the build command
```

but `build` itself has no way to pass parameters to the corresponding build recipes (in this case `build-recipe-spec`, see https://github.com/openSUSE/obs-build/blob/7950ca8f72622d2f33b141c83184257094a5ed08/build-recipe-spec#L170)

rhabacker:
Fixes #922

### DIFF
--- a/build
+++ b/build
@@ -437,7 +437,10 @@ Known Parameters:
   --norootforbuild
               Force building with user 'abuild', ignore the "needsrootforbuild"
               hint from the recipe file
-              
+
+  --buildtool-opt=OPTIONS
+              Pass OPTIONS to the used build tool (currently supported: rpmbuild)
+
 
 Remember to have fun!
 
--- a/build-recipe
+++ b/build-recipe
@@ -68,6 +68,11 @@ recipe_cleanup () {
 
 recipe_parse_options() {
     case ${PARAM/#--/-} in
+      -buildtool-opt)
+	needarg
+	BUILD_TOOL_OPT="$ARG"
+	shift
+	;;
       -stage)
 	needarg
 	BUILD_RPM_BUILD_STAGE="$ARG"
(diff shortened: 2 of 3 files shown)


rhabacker:
Someone here how can merge these pr?  

There is a request to remove the deprecated dependency generator from rpmbuild, see https://github.com/rpm-software-management/rpm/issues/2373.

The removal of the deprecated generator is underway, see https://bugzilla.opensuse.org/show_bug.cgi?id=1175587, but for finding and fixing bugs related to the new generator, it would be helpful if package maintainer could simply, e.g. 

    osc build --build-opt="--recipe-opt=--rpmfcdebug" openSUSE_Leap_15.4

to debug that.

adrianschroeter:
We don't really like the "--receipe-opt" name as it does not affect the build receipt, but the processing tool.

What about calling it --buildtool-opt or --engine-opt and support that also in osc directly to avoid the need to cascade via --build-opt there?

rhabacker:
> support that also in osc directly to avoid the need to cascade via --build-opt there?

I think this can only happen after the next version of the `build` package, as it does not currently provide for the return of the current version.
```
build --version
Unknown option '--version'. Exit.
```
But this would be necessary to distinguish older version, otherwise an error message like "Unknown option --buildtool-opt..." would appear when called.

rhabacker:
Since build does not currently support a `--version` command line parameter, a workaround would be to call `build --help` and look for the required `--buildtool-opt` option to see if it is available.

PR #929, opened 2023-04-24
config: sync tumbleweed config and move to suse_version 1699

DimStar77:


DimStar77:
openSUSE:Factory:Staging:A is already setup to use suse_version 1699 - I hope that's tall the changes needed (and that I did not mess up the config )

PR #931, opened 2023-04-26
Remove blank paragraphs and regenerate the html

crazyss:
Some of BuildFlags is missing in HTML.

4.6 Build Config chapter is redundant.

4.5 Repository setup seems like missing context so remove it.


crazyss:
@mlschroe ,please take a look.

PR #932, opened 2023-05-05
Fix startdockerd with cgroup v1

Vogtinator:
If cgroup v1 is used, neither /sys/fs/cgroup/unified nor /sys/fs/cgroup/cpu.stat exist, so it goes into the fallback path to mount the cgroup hierarchy manually. This is mounted on top of the actual cgroup mounts, breaking them. Fix this by recognizing a working cgroupv1 mount.


Vogtinator:
FWICT `mount -t cgroup cgroup /sys/fs/cgroup -o devices` is also broken and can't work, but that's another topic...

PR #934, opened 2023-05-17
Initial OpenWrt setup

ndren:
This is an attempt to get set up for OpenWrt builds for GSoC 2023.

The code and comments are mostly unnecessary (noop) at the moment and the plan is to work on them to get a clean setup working. Currently I'd like to know if the code design is accurate. For example, is running `make download` a reasonable approach to getting build dependencies installed? (It is the official way to do this under OpenWrt, and does not compile anything, only downloads files.)

### DIFF
--- a/Build/OpenWrt.pm
+++ b/Build/OpenWrt.pm
@@ -1,0 +1,184 @@
+################################################################
+#
+# Copyright (c) 2017 SUSE Linux Products GmbH
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program (see the file COPYING); if not, write to the
+# Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+#
+################################################################
+
+package Build::OpenWrt;
+
+use strict;
+use warnings;
+use Build::Deb;
+use Build::Rpm;
+
+my $yamlxs = eval { require YAML::XS; $YAML::XS::LoadBlessed = 0; return 1 };
+my $yamlpp = eval { require YAML::PP; return YAML::PP->new };
+
+sub _have_yaml_parser {
+  return $yamlpp || $yamlxs ? 1 : undef;
+}
+
+sub _load_yaml {
+  my ($yaml) = @_;
+  my $data;
+  if ($yamlpp) {
+    $data = eval { $yamlpp->load_string($yaml) };
+    return $data;
+  }
+  if ($yamlxs) {
+    eval { $data = YAML::XS::Load($yaml) };
+    return $data;
+  }
+  die "Neither YAML::PP nor YAML::XS available\n";
+}
+
+sub _load_yaml_file {
+  my ($fn) = @_;
+  my $data;
+  if ($yamlpp) {
+    $data = eval { $yamlpp->load_file($fn) };
+    return $data;
+  }
+  if ($yamlxs) {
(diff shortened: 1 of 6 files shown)


ndren:
This PR is ready for review, let me know what you think. Thanks!

Build/OpenWrt.pm
```diff
@@ -0,0 +1,150 @@
...
+    return { error => "Failed to parse JSON file '$fn'" } unless defined $data;
+  } elsif (ref($fn) eq 'SCALAR') {
+    $data = _load_yaml($$fn);		# used in the unit test
+    return { error => "Failed to parse '$fn'" } unless defined $data;
+  } else {
+    $data = _load_yaml_file($fn);
+    return { error => "Failed to parse file '$fn'" } unless defined $data;
+  }
+  return $data;
+}
```
dcermak, line 81: I'm pretty certain you don't need this part, or at least not all of it. You should be fine with a json parser.

build-recipe-openwrt
```diff
@@ -0,0 +1,128 @@
+#################################################################
+#
+# OpenWrt specific functions.
+#
+# Adapted from Flatpak specific functions, see details below.
+#
+# Author: Tina Müller <tina.mueller@suse.com>
```
dcermak, line 7: This one is wrong ;-)

dcermak, line 7: You might want to change this part, at least the author is incorrect ;-)

build-recipe-openwrt
```diff
@@ -0,0 +1,128 @@
...
+        git clone "$URL" "$BUILD_ROOT/git"
+        sed -i "s|$URL|/git|" "$BUILD_ROOT/repo/$NAME/Makefile"
+    fi
+
+
+    chroot "$BUILD_ROOT" bash -c "cd sdk && grep localrepo feeds.conf.default || echo \"src-link localrepo /repo/\" >> feeds.conf.default"
+
+
+    # Download the dependencies
+    chroot "$BUILD_ROOT" /bin/bash -c "cd sdk && make download V=s"
```
dcermak, line 98: This will fail on OBS where networking is disabled.

Build/OpenWrt.pm
ndren: That makes sense, I wasn't sure whether a JSON parser was present inside the build system. I saw https://github.com/openSUSE/obs-build/blob/master/Build/SimpleJSON.pm but wasn't sure if that would work, and couldn't verify whether https://metacpan.org/dist/JSON was preinstalled.

build-recipe-openwrt
ndren: Thanks, let me know if this works.

Build/OpenWrt.pm
ndren: The nice part about this method is that it keeps yaml parsing available if someone (or something, e.g. a buildsystem) prefers YAML.

build-recipe-openwrt
ndren: I am still wondering what I need to do about build-time dependencies (`make package/$NAME/download`), do I assume those are passed in as build inputs somehow?

ndren: I'm currently thinking of maybe having a list of git repos that are copied in that are comitted in the source build tree, though I'm not sure how that interacts with package/$NAME/download.

ndren: I have a question -- do you know if I'm able to use https://github.com/openSUSE/obs-build/blob/master/PBuild/RemoteAssets.pm before building?

adrianschroeter: depends, pbuild is executing it always. 

For OBS scenarios the "download_asset" service would need to get enabled for that package or project.

adrianschroeter: regarding dependencies, you may define a default as     build-packages:openwrt     in Build.pm in %subst_defaults.

This would require these packages by default (but can be overwritten by the user). 

You may also define an openwrt config file where you define default repository/ies via RepoURL: tag

Build/OpenWrt.pm
```diff
@@ -0,0 +1,150 @@
+################################################################
+#
+# Copyright (c) 2017 SUSE Linux Products GmbH
```
adrianschroeter, line 3: you are of course allowed to use/add your name here ;)

PR #938, opened 2023-06-19
configs/alp: add is_alp macro

danishprakash:
One of our use cases can do away with some of the complexity if `is_alp` were to be supported. If this is not planned for some reason, is there a way to identify builds on ALP and otherwise? Thanks


dcermak (approved):

Vogtinator:
You could just use `%if 0%{suse_version} >= 1600 && !0%{?is_opensuse}` instead

dirkmueller:
I think currently "ALP" is `== 1600`. I am not sure we want to have a "is_alp" macro everywhere before the tracks and versioning scheme is clarified. I'm leaning towards not merging that until that is clearer (just to prevent use)

danishprakash:
Thanks for the help, the above suggestions should do the job. Closing this PR.

PR #944, opened 2023-07-14
Pass-through --buildtool-opt

bmwiedemann:
Pass-through `--buildtool-opt` into VMs to fix

    osc build --vm-type=kvm --build-opt=--buildtool-opt=--noclean


adrianschroeter:
clean happens outside of the vm, why is that needed inside VM?

bmwiedemann:
This is about `rpmbuild` that now removes the BUILD dir after a successful build.
https://github.com/rpm-software-management/rpm/commit/b34333fa021c0ee7215714eeef96d1a2843ea08e

PR #945, opened 2023-07-25
[Helm] Add logic to define BUILD_PATH based on '_service' file

arsenalpoll:
This commit adds a block of code that checks if the _service directory exists. If it does, it parses the _service file to extract the `filename` and `extract` parameters. It then uses these values to define the `BUILD_PATH`.

### DIFF
--- a/build-recipe-helm
+++ b/build-recipe-helm
@@ -60,6 +60,19 @@ recipe_prepare_helm() {
 }
 
 recipe_build_helm() {
+    BUILD_PATH=
+    if [ -d "$BUILD_ROOT/$TOPDIR/BUILD/_service" ]; then
+      # get the packaging file name
+      FILENAME=$(cat $BUILD_ROOT/$TOPDIR/BUILD/_service | xmllint --xpath 'string(//param[@name="filename"])' -)
+      # get specified path
+      EXTRACT=$(cat $BUILD_ROOT/$TOPDIR/BUILD/_service| xmllint --xpath 'string(//param[@name="extract"])' -)
+      # define $BUILD_PATH
+      if [ -z "$EXTRACT" ]; then
+        BUILD_PATH="$FILENAME/"
+      else
+        BUILD_PATH="$FILENAME/$(dirname "$EXTRACT")"
+      fi
+    fi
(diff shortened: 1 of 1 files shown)


adrianschroeter:
A source directory in build area that you parse as xml????

Sorry, this looks highly suspicious, did you execute it once at least?

arsenalpoll:
> A source directory in build area that you parse as xml????
> 
> Sorry, this looks highly suspicious, did you execute it once at least?


@adrianschroeter  Hi, thank you very much for your reply.

This is my use example:
```xml
<services>
  <service name="obs_scm">
      <param name="scm">git</param>
      <param name="url">http://gitlab.xx.xx/abc/xyz.git</param>
      <param name="version">_none_</param>
      <param name="revision">origin/dev</param>
      <param name="filename">abc_xyz</param>
      <param name="extract">build/charts/crd/Chart.yaml</param>    
  </service>
</services>
```

`Chart.yaml` is extracted from the Git storage repository. 
`helm package` command needs to be executed within the path where the Chart.yaml is located, so I added the specified path.

My example helm chart.tgz directory tree:
```shell
[root@arsenal ]# tree abc_xyz
abc_xyz
├── Chart.yaml
├── templates
│   ├── k8s.ovn.org_routes.yaml
│   ├── network-crd.yaml
│   └── subnet-crd.yaml
└── values.yaml
```

Content under the git repository path `build/charts/crd/` 
image

build-recipe-helm
```diff
@@ -60,6 +60,19 @@ recipe_prepare_helm() {
 }
 
 recipe_build_helm() {
+    BUILD_PATH=
+    if [ -d "$BUILD_ROOT/$TOPDIR/BUILD/_service" ]; then
+      # get the packaging file name
+      FILENAME=$(cat $BUILD_ROOT/$TOPDIR/BUILD/_service | xmllint --xpath 'string(//param[@name="filename"])' -)
```
adrianschroeter, line 66: you can not create a check that _service is a directory and afterwards reading it as an XML file.

also the "cat" command is quite pointless, you could add the file name to xmllint directly.

But what worries me most is that there should not be a source file in the BUILD directory at all, no?

PR #957, opened 2023-10-06
Unpack zstd using --zstd option to tar, not -J

andrewshadura:
Pass a correct argument to tar to unpack zstd. This apparently doesn’t matter to libarchive/bsdtar which detects the compression anyway, or to the unpackarchive tool, but let’s use the right option for the correctness sake.


andrewshadura:
Right, having *actually* checked it, it seems bsdtar doesn’t accept `--zstd` 😢

PR #963, opened 2023-11-08
generate_sbom: fix support for older RPM versions

jgleissner:
Do not abort in dump_rpmdb in case rpmdb does not support --exportdb but use fallback using /usr/lib/rpm/rpmdb_dump.

This fixes SBOM generation on SLE 12 build hosts.


adrianschroeter:
ah, sorry, did not catch this pull request. Just looked into it and fixed it as well, the change is live.

PR #970, opened 2023-12-18
Add support for setting the ccache archive size

dcermak:


adrianschroeter:
This defines a default limit of 5GB.  This would disable it for nodejs-electron in Factory. Not sure if we want this?
@DimStar77 @AdamMajer Do you have an opinion?

@dcermak What is your rational, where do you think it should be disabled by default?

Note, it won't be possible to overwrite this again in OBS.

dcermak:
> @dcermak What is your rational, where do you think it should be disabled by default?

Huh, I thought this would just use the default of 5GB if unset?

> Note, it won't be possible to overwrite this again in OBS.

I don't really mind that. I just wanted to have this feature when building a very large C/C++ package locally that would fill the ccache and then wouldn't benefit that much from having a cache.

AdamMajer:
```
> cat .ccache/ccache.conf 
max_size = 5.0G
```

I don't remember setting this. Looking at TFM (ccache manpage)

> 
>  max_size (CCACHE_MAXSIZE)
>  This option specifies the maximum size of the cache. Use 0 for no limit. The default value is 5G. Available suffixes: k, M, G, T (decimal) and Ki, Mi, Gi, Ti (binary). The
>  default suffix is G. See also CACHE SIZE MANAGEMENT.
> 


> SIZE MANAGEMENT
>        By default, ccache has a 5 GB limit on the total size of files in the cache and no limit on the number of files. You can set different limits using the command line options
>        -M/--max-size and -F/--max-files. Use the -s/--show-stats option to see the cache size and the currently configured limits (in addition to other various statistics).
> 
> Cleanup can be triggered in two different ways: automatic and manual.

which means this PR is unnecessary and ccache is accidentally overflowing for `nodejs-electron` already

adrianschroeter:
k, closing it

dcermak:
I beg to object, this is still useful for local builds.

PR #982, opened 2024-02-19
Pass release for reproducible builds when testing

JanZerebecki:
also enabling changelog handling and move changelog to changes file, to
more closely do what OBS does normally.

Passing a release might become required in the future on distributions
that have reproducible builds enabled, as otherwise
SOURCE_DATE_EPOCH_MTIME can not be computed. rpms build without that
would result in a hard to notice failures of tools that rely on mtimes
like rsync without --checksum.

### DIFF
--- a/test/common
+++ b/test/common
@@ -101,6 +101,7 @@ run_build()
 	set -- $linux32 $SU_WRAPPER \
 		$BUILD_DIR/build \
 		--root "${build_root}" \
+		--changelog --release 5.23 \
 		"${repos[@]}" \
 		"${build_args[@]}"
 	echo "$@"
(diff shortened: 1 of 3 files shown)


JanZerebecki:
CI failure is false positive, the blank line is necessary. One could disable that check by replacing the invocation with: `git --no-pager -c core.whitespace=-blank-at-eof diff --check origin/master..HEAD --`

JanZerebecki:
rebased on top of https://github.com/openSUSE/obs-build/pull/981

JanZerebecki:
On second thought it would be better if osc build worked without a release argument. Not sure yet, how to make osc build succeed and obs fail without a release...

PR #983, opened 2024-02-27
Show rpm Requires and Provides

dmach:


dmach:
I needed to display rpm deps during the build to understand why a package got pulled in, this is an improved version of my hack. We should probably enable this only when an option or env variable is on.

dmach:
Fixed the root cause instead: https://github.com/openSUSE/osc/pull/1500 (obs:cli_debug_packages were not pulled in by default)

PR #987, opened 2024-03-12
docker: decode ascii encoded dockerarg strings

adrianschroeter:
To support for example spaces via %20

Alternative implementation for pull request 965


adrianschroeter:
Alternative for https://github.com/openSUSE/obs-build/pull/965

PR #997, opened 2024-04-16
Adding debian support to generate_sbom

cschneemann:
Added support for debian images to generate_sbom.
It checks if it is a debian based image by evaluating the os-release file.

It introduces some new BuildFlags to configure "special" behaviour:

`spdx-declared-license:NOASSERTION` licenseDeclared in spdx info for a package is set to "NOASSERATION",
if not set the found license from the package information is used (like for "licenseConcluded").

The license information is taken from the package's copyrightfile.

The generation of spdx' file section can be disabled, if just the package information is wanted.
This can be switched by buildflag "spdx-files-generation". To switch it off it has to be set to "no":
`spdx-files-generation:no`.
Default is to generate also the files information. This option is just available in the deb-handling. No changes to rpm made here.

As debian packages does not have a vendor information this information is taken from the
OBS vendor macro/OBS instance name.

Added the sbom generation also to build-receipe-livebuild for a first usage.

### DIFF
--- a/build-recipe-livebuild
+++ b/build-recipe-livebuild
@@ -264,14 +264,20 @@ recipe_build_livebuild() {
 	done
     done
 
+    BASE_NAME="${RECIPEFILE%.livebuild}-${ARCH}${buildnum}"
     for format in $(queryconfig --dist "$BUILD_DIST" --configdir "$CONFIG_DIR" --archpath "$BUILD_ARCH" buildflags+ sbom | sort -u) ; do
         echo "Generating $format sbom file"
 	    pushd $BUILD_ROOT/$TOPDIR/$LIVEBUILD_ROOT >/dev/null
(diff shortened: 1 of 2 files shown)


mlschroe:
What's with the "just join licenses" commit? Why is it needed? I think it was correct before, not there is `$ret{'license'} = @license_uniq[0];` which doesn't seem to be good perl...

cschneemann:
> What's with the "just join licenses" commit? Why is it needed? I think it was correct before, now there is `$ret{'license'} = @license_uniq[0];` which doesn't seem to be good perl...

The copyrightfile for  package can has multiple licenses in it. Every found license is pushed into @license. This gets reduced by removing duplicate entries. So from many licenses it can be stripped to just one license in the array and then the entry in the file would be
`"licenseConcluded": " AND license",`
With this commit the last/only license in the array is written into $ret{'license'}.

cschneemann:
> What's with the "just join licenses" commit? Why is it needed? I think it was correct before, now there is `$ret{'license'} = @license_uniq[0];` which doesn't seem to be good perl...

deleted my last comment, because it was "stupid"..  The behaviour I wanted to fix with this was fixed by another change.

PR #999, opened 2024-04-22
pbuild: Handle string instead of hashref

bmwiedemann:
Without this patch, pbuild (of openSUSE:Factory:Rings:0-Bootstrap standard x86_64) failed with

    building binutils/binutils.spec
    Can't use string ("binutils") as a HASH ref while "strict refs" in use at /usr/lib/build/PBuild/Checker.pm line 731.


mlschroe:
Fixed in a different way.

bmwiedemann:
linking commit 5a078c2b0f6904060e25621a263fd2a14c3c04e0 to make it easier to find.

PR #1002, opened 2024-05-09
Drop automatic inclusion of .pc file for baselibs.

### LINKED ISSUE #724: Have 32bit choice when build with pkgconfig
When I build rpm with build script, if my spec has pkgconfig something, build will say have choice for pkgconfig (): ... ...-32bit
An example is:
When I build grub2
My command is 
sudo build --debug --repo https://mirror.sjtu.edu.cn/opensuse/update/leap/15.3/sle/ --repo https://mirror.sjtu.edu.cn/opensuse/update/leap/15.3/backports/ --repo https://mirror.sjtu.edu.cn/opensuse/update/leap/15.3/leap/ --repo https://mirror.sjtu.edu.cn/opensuse/update/leap/15.3/oss/ --repo https://mirrors.huaweicloud.com/opensuse/distribution/leap/15.3/repo/oss/ --dist sle15.3 grub2-2.04-20.4.src.rpm
It will tell me
expansion error
have choice for pkgconfig (libudev) needed by device-mapper-devel: libudev-devel libudev-devel-32bit
have choice for pkgconfig (bzip2) needed by freetype2-devel: libbz2-devel libbz2-devel-32bit
I know maybe it is not a bug, but could you tell my how to resolve it?

badshah400:
Fixes #724.

Avoid default inclusion of pkgconfig files in automatically generated baselibs configured biarch packages. This avoids the "have choice" issue seen with `BuildRequires: pkgconfig(FOO)` type dependencies on OBS for any dependents of a biarch FOO-devel package. It is wrong to ask every project supporting biarch devel packages to configure project macros to disfavour one pkgconfig based dependency in favour of another.

Packager may at their wish explicitly include the pkgconfig file in the biarch package explicitly by appropriately configure the associated baselibs.conf file, but as a default, this should be hardly ever required.

### DIFF
--- a/baselibs_configs/baselibs_global.conf
+++ b/baselibs_configs/baselibs_global.conf
@@ -24,7 +24,7 @@ targettype 64bit extension 64
 
 targetname <name>-<targettype>
 
-+.*/lib(64|ilp32)?/.*\.(so\..*|so|o|a|la|pc)$
++.*/lib(64|ilp32)?/.*\.(so\..*|so|o|a|la)$
 
 targettype 64bit -^(/usr)?/lib(ilp32)?/lib
 targettype 32bit -/lib64/


badshah400:
Ping?

mlschroe:
See https://bugzilla.suse.com/show_bug.cgi?id=1172563, which asked for adding the .pc files.

badshah400:
Off the top of my head, the other option would be to turn off `AutoProv` for baslibs generated packages that do install .pc files.

DimStar77:
If OBS could handle Prefer using RegEx, one could do:

`Prefer: -.*-devel-32bit` and be done for all of the -devel-32bit packages

badshah400:
> If OBS could handle Prefer using RegEx, one could do:
> 
> `Prefer: -.*-devel-32bit` and be done for all of the -devel-32bit packages

That would also have to be done for every project, wouldn't it?

DimStar77:
> That would also have to be done for every project, wouldn't it?

yes (let alone that OBS would actually need to support it first); but we only have a list of like 3 supported distros- so adding this in 3 places would not be really hard

PR #1008, opened 2024-06-07
config: sync Tumbleweed config

DimStar77:
Change needed to address this new test failure:

https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:A/build/standard/x86_64

gettext-runtime is splitting out envsubst (andenvsubst-mini); the mini flavor does, as usual, require `thisis-only-for-build-envs' which is to be ignored by OBS/build


ana:
Hi @mlschroe @adrianschroeter 
Could you please take a look so we can update gettext-runtime in Factory? Thank you!

bugfinder (approved):

PR #1009, opened 2024-06-12
fix: handle trailing slashes in BUILD_ROOT path comparison

myml:
In the current implementation, we compare the MYSRCDIR variable with a constructed path using the BUILD_ROOT variable. This comparison can fail if BUILD_ROOT ends with a trailing slash(eg. In KVM, BUILD_ROOT is set to '/'), causing copy behavior.


adrianschroeter:
hm, we should do this in a single place IMHO. best would be at the argument parsing in build itself

mlschroe:
Isn't that no longer needed as of commit dca0e07939ad80f2920d6e771447687aebdfe1bc ?

PR #1011, opened 2024-07-01
Tumbleweed config: enable preinstallexpand

DimStar77:


DimStar77:
This will help us get over
https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:C/build/standard/x86_64

the flag has been tested in that staging and is confirmed to not 'explode' our cycle

In a next step I will start cleaning up the 'Preinstall' (and VMInstall) lines

PR #1017, opened 2024-08-01
Fix invalid type '\' in unpack

andreas-schwab:


adrianschroeter:
mls fixed it meanwhile

PR #1018, opened 2024-08-02
exclude unsupported arch for Arch build

YukariChiba:
Currently Arch build script does not filter out unsupported architectures in `arch=()` and set `exclarch`.

According to [ArchLinux Ports RFC](https://gitlab.archlinux.org/dvzrv/rfcs/-/blob/rfc/arch_linux_ports/rfcs/0032-arch-linux-ports.rst), the `arch` field for PKGBUILDs is now not limited to x86 architectures such as x86_64 i386

### DIFF
--- a/Build/Arch.pm
+++ b/Build/Arch.pm
@@ -159,6 +159,13 @@ sub parse {
     my @assets = get_assets(\%vars, $asuf);
     push @{$ret->{'remoteassets'}}, @assets if @assets;
   }
+  my @exclarch;
+  push @exclarch, @{$vars{'arch'}};
+  @exclarch = grep {$_ ne "any"} @exclarch;
+  # unify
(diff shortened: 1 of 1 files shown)


Build/Arch.pm
```diff
@@ -159,6 +159,13 @@ sub parse {
     my @assets = get_assets(\%vars, $asuf);
     push @{$ret->{'remoteassets'}}, @assets if @assets;
   }
+  my @exclarch;
+  push @exclarch, @{$vars{'arch'}};
+  @exclarch = grep {!/any/} @exclarch;
```
adrianschroeter, line 164: this would also hit any possible arch containing "any" string. May not be a problem right now, but better write this as    {$_ ne "any"}    instead

adrianschroeter (changes requested):

PR #1021, opened 2024-08-13
Save container in oci format so that oci-defined annotations are

dirkmueller:
preserved


mlschroe:
Problem is it needs to round-trip through OBS.

mlschroe:
(Also there is an "else" branch missing so that it saves the container in the podman case if the format is not set to oci...)

dirkmueller:
> (Also there is an "else" branch missing so that it saves the container in the podman case if the format is not set to oci...)

true, that's embarassing.

PR #1022, opened 2024-08-14
pbuild: Use the Release value from _config

bmwiedemann:
For reproducible-builds I would like
to be able to produce identical binaries to OBS.

In prjconf I added `Release: 1.1`
but without this patch, pbuild ignored this.

We substitute the OBS CI_CNT and B_CNT counters with 1 to remain compatible.

Because there is a default value of `<CI_CNT>.<B_CNT>`, this will change the default package names from `foo-123-0` to `foo-123-1.1`

This patch was done while working on reproducible builds for openSUSE, sponsored by the NLnet NGI0 fund.


bmwiedemann:
solved differently with f47374bc2027b3a21c2232e3d88b2b296ddd9483

PR #1023, opened 2024-08-21
dist: add more required recommendations for KVM builds

jengelh:
A worker system may have obs-worker installed, but because /usr/bin/build is transferred from obsrepserver, the dependencies of obs-build are not carried over. KVM builds thus fail complaining about absence of mkfs.ext4, mkfs.xfs or the hostname utility (especially if one used zypper in --no-recommends obs-worker, but that's another story).


adrianschroeter:
however, these should be added to obs-worker as well to have an effect there.

adrianschroeter:
Solving that via a recommend in obs-worker: https://github.com/openSUSE/open-build-service/pull/16771

PR #1036, opened 2024-11-04
Add armv7hl to build-vm-qemu

stoecker:
When I use "osc build" on my local machine cross building for 32 bit ARM I always get an issue because no tool is called instead of the ARM toolchain.

For approx a year I fix the build-vm-qemu after each update hoping that it will be fixed upstream.  As it was not fixed I now searched the original repo (BTW it would be helpful when the "build" package of openSUSE actually contained a link to this repo).

This pull request fixes the issue.

### DIFF
--- a/build-vm-qemu
+++ b/build-vm-qemu
@@ -38,7 +38,7 @@ vm_verify_options_qemu() {
 vm_startup_qemu() {
     # overwrite some options for specific build architectures
     case $BUILD_HOST_ARCH in
-        armv6l|armv7l)
+        armv6l|armv7l|armv7hl)
             qemu_bin="/usr/bin/qemu-system-arm"
             qemu_console=${qemu_console:-ttyAMA0}
             qemu_options="-M virt"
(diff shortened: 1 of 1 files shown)


adrianschroeter:
The hostarch exists only as armv7l since there is only armv7l hardware and a linux kernel.

armv7hl is a userland decision, so you need some

Target: armv7hl-...

definition. Or do I miss some point here?
Do you have an example setup?

stoecker:
I have no idea why it is necessary, but is needed for every ARM 32bit repository I use. This is also needed on other machines my colleagues operate which have been setup independend, so it's no local configuration issue.

Example:
```
  <repository name="openSUSE_Leap_15.5_ARM">
    <path project="openSUSE:Leap:15.5:ARM" repository="ports"/>
    <arch>armv7l</arch>
  </repository>
```

I didn't do any other settings, special targets or whatever. Simply build for 32bit ARM on an x64_64 bit system.

Running with
osc build --local-package -j 4 --vm-disk-size=8096 --vm-type=qemu ...

stoecker:
Any reason why this is not accepted?
- It solves a real issue
- In no situation it can do any harm
- After 5 months it is clear that you don't have a better solution.

PR #1037, opened 2024-11-18
Add support to parse build params from _buildparams file

bmwiedemann:
To allow for custom build tweaks such as `-j1` and
    `--vm-custom-opt=-cpu qemu64` written as separate lines in the `_buildparams` file.

This includes regexp whitelist validation to minimize risk to OBS operations.
    
The pattern for jobs deliberately does not have a `*` or `+` to not allow DoS or integer overflows.

The pattern for CPU allows to enable or disable individual CPUID flags. IMHO this should be safe.


adrianschroeter:
After some discussion, a possible way to implement this could be via buildflags. eg.

#!BuildFlags: cputype:$KNOWN_GOOD_CPU_TYPE

the known good cpu type may cause issues when eg. an AMD cpu gets enforced on a Intel hardware? But at least cpu types like "qemu64" on x86_64 could be supported here.

Alternative/Additionally we may support

#!BuildFlags: cpuflag:$FLAG

FLAG can be for example "-avx" to disable avx support.

This way we can have architecture specific definitions, because these can be sourounded via %ifarch statements.

mlschroe:
To be honest, I don't like this at all. Which packages would need this at all? If this is a very small list, can't that be fixed in the packages in a different and more sane way?

bmwiedemann:
There is a small list:
* [colord](https://bugzilla.opensuse.org/show_bug.cgi?id=1217747) (AVX2)
* [calibre](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/67ICWGGE3TUPG5RH32GZAXICO4T5BXFG/) (SSE-4.1)
* [emacs](https://bugzilla.suse.com/show_bug.cgi?id=1239628) (very sensitive to CPU changes)
* (not counting the dozen packages that used `-march=native` that we already patched)

These are still left after quite some time.

PR #1048, opened 2025-01-05
mkosi: drop most systemd/build-packages deps and use obs_scm directory as source if present

bluca:
The package itself can pull in what it needs. The list of dependencies changes too often to be hard-coded here.
Also allow storing recipes in git via obs_scm.

### DIFF
--- a/build-recipe-mkosi
+++ b/build-recipe-mkosi
@@ -89,6 +89,16 @@ recipe_build_mkosi() {
         fi
     fi
 
+    # If we are running with any source services, the sources might be unpacked in a directory
+    # in SOURCES/, so try to find the right one, by checking where the top level mkosi.conf is.
+    workdir="$TOPDIR/SOURCES/"
+    for d in "$TOPDIR"/SOURCES/*; do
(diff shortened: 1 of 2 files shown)


build-recipe-mkosi
```diff
@@ -89,6 +89,16 @@ recipe_build_mkosi() {
         fi
     fi
 
+    # If we are running with the obs_scm service, the sources will be unpacked in a directory
+    # in SOURCES/, so try to find the right one, by checking where the top level mkosi.conf is.
+    workdir="$TOPDIR/SOURCES/"
+    for d in "$TOPDIR"/SOURCES/*.obsinfo; do
+        if [ ! -d "${d%.obsinfo}" ]; then continue; fi
+        if [ ! -f "${d%.obsinfo}/mkosi.conf" ]; then continue; fi
+        workdir="${d%.obsinfo}"
```
adrianschroeter, line 98: please do not relay on .obsinfo files. You can also have directories via other mechanics. It will esp. break when using native git.

Just look for any directory with a mkosi.conf.

adrianschroeter (changes requested):

PR #1053, opened 2025-01-27
config: sync Tumbleweed config

DimStar77:
This should fix the currently seen build fail in
https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:N/build/standard/x86_64

prjconf needed an extra Prefer: for ncurses

(includes 6 months worth of prjconf changes; the ncurses change is just the one that forced the sync to happen)

### DIFF
--- a/configs/tumbleweed.conf
+++ b/configs/tumbleweed.conf
@@ -1,5 +1,5 @@
 Prefer: -yast2-schema-micro -yast2-schema
-%define gcc_version 13
+%define gcc_version 14
 
 # testing new cycle algorithm
 BuildFlags: genmetaalgo:1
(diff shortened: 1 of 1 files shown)


bitstreamout (changes requested):
Please change line 626 that it becomes:

```
# have choice for libtinfo.so.6()(64bit) needed by libreadline8: libncurses6 libncurses6-compat
Prefer: libncurses6
```

DimStar77:
> Please change line 626 that it becomes:
> 
> ```
> # have choice for libtinfo.so.6()(64bit) needed by libreadline8: libncurses6 libncurses6-compat
> Prefer: libncurses6
> ```

Fixed! (using the de-prefer mechanism as we do in most cases, i.e `Prefer: -libncurses6-compat`)

bitstreamout (approved):
That should work now.

PR #1054, opened 2025-01-28
Drop needsrootforbuild

bmwiedemann:
Drop `needsrootforbuild`
It seems to not be needed anymore.
It was part of the initial git import in 2017.


adrianschroeter:
it is used for the test build for verification of the delivered code and target configuration.

PR #1057, opened 2025-02-07
Support xz compressed Debian repos

adrianschroeter:


adrianschroeter:
jfyi, @lnussel I imported your mail on request by @mlschroe

mlschroe:
Note that the OBS code does the "right" thing and downloads the Release file. It then uses xz if available. We should do the same in pbuild. Dunno if we need this patch after that change.

PR #1058, opened 2025-02-13
mkosi: add back system-packages used by build-recipe directly

bluca:
The previous commit was a bit too zealous in dropping system-packages. Add back those that are used directly by the obs-build build-recipe-mkosi as the mkosi packages in the distros won't necessarily use them and thus depend on them directly.

Follow-up for 69e46e72f4be385bd428e4656b04a70395ad73ea

### DIFF
--- a/Build.pm
+++ b/Build.pm
@@ -159,13 +159,13 @@ my %subst_defaults = (
     'apt-utils', 'cpio', 'dpkg-dev', 'live-build', 'lsb-release', 'tar',
   ],
   'system-packages:mkosi:rpm' => [
-    'mkosi',
+    'mkosi', 'createrepo', 'gzip',
   ],
   'system-packages:mkosi:deb' => [
-    'mkosi',
+    'mkosi', 'dpkg-dev', 'lsb-release', 'gzip',
   ],
   'system-packages:mkosi:arch' => [
-    'mkosi',
+    'mkosi', 'arch-install-scripts', 'gzip',
   ],
   'system-packages:mock' => [
     'mock', 'system-packages:repo-creation',


Build.pm
```diff
@@ -159,13 +159,13 @@ my %subst_defaults = (
...
-    'mkosi',
+    'mkosi', 'createrepo', 'gzip',
   ],
   'system-packages:mkosi:deb' => [
-    'mkosi',
+    'mkosi', 'dpkg-dev', 'lsb-release', 'gzip',
   ],
   'system-packages:mkosi:arch' => [
-    'mkosi',
+    'mkosi', 'arch-install-scripts', 'gzip',
```
DaanDeMeyer, line 168: `arch-installs-scripts` can be dropped, we don't use it anymore in mkosi and haven't for a very long time.

Build.pm
```diff
@@ -159,13 +159,13 @@ my %subst_defaults = (
     'apt-utils', 'cpio', 'dpkg-dev', 'live-build', 'lsb-release', 'tar',
   ],
   'system-packages:mkosi:rpm' => [
-    'mkosi',
+    'mkosi', 'createrepo', 'gzip',
   ],
   'system-packages:mkosi:deb' => [
-    'mkosi',
+    'mkosi', 'dpkg-dev', 'lsb-release', 'gzip',
```
DaanDeMeyer, line 165: Same here, `dpkg-dev` isn't really useful to mkosi without having apt as well. I'm a bit confused what these deps are for since building debian/ubuntu/arch from a opensuse tumbleweed doesn't work at all because the necessary tools are not packaged in opensuse.

Build.pm
bluca: This is not about what mkosi uses, but what the glue script in this repo uses. mkosi can pull in its own dependencies, but this needs to pull in what the script uses itself, so that it's guaranteed to be there

Build.pm
bluca: As above, this is what will be pulled in for debian-based image builds and that the glue from this repo itself will use, the mkosi package can pull in what it needs by itself

PR #1062, opened 2025-02-24
POSIX sh compat

jengelh:


adrianschroeter:
the /bin/bash change would break busybox support, but it just got fixed different for cases like /bin/sh to dash (ubuntu).

Could you rebase and drop that commit?

adrianschroeter:
"mkbaselibs: modernize generated specfiles and add sh quoting"

I am told that this would break reproducible builds

Can you drop that one as well?

jengelh:
> the /bin/bash change would break busybox support

But once we request /bin/bash, busybox is not even involved.
Anyhow, it was already broken; busybox-ash only gained ksh-style `<(cmd)` support after SLE 15 GA.

> I am told that this would break reproducible builds

Objection, hearsay. But also, I would love to have more substance than just "this breaks"; breaks how.

adrianschroeter:
with busybox I mean minimal containers where only busybox is available and no bash.

Dunno if is still working I have to admit, but for build script we need to support also old builds. (would be great if we could also fix current distros indeed).

adrianschroeter:
however, I pushed yesterday an alternative solution where we try to use bash when available. This should be fully backward compatible and still use bash when available.

jengelh:
Still, mkbaselibs (updated) still needs the rest. This should resolve errors seen in openSUSE:Factory:Staging:gcc7/flac and everything that uses baselibs.

```
[  291s] Executing(%install): /bin/dash -e /var/tmp/rpm-tmp.SodtsW
[  291s] + umask 022
[  291s] + cd /home/abuild/rpmbuild/BUILD/flac-1.5.0-build
[  291s] + /usr/bin/rm -rf /home/abuild/rpmbuild/BUILD/flac-1.5.0-build/BUILDROOT
[  291s] + /usr/bin/mkdir -p /home/abuild/rpmbuild/BUILD/flac-1.5.0-build
[  291s] + /usr/bin/mkdir /home/abuild/rpmbuild/BUILD/flac-1.5.0-build/BUILDROOT
[  291s] + rm -rf /home/abuild/rpmbuild/BUILD/flac-1.5.0-build/BUILDROOT
[  291s] + mkdir /home/abuild/rpmbuild/BUILD/flac-1.5.0-build/BUILDROOT
[  291s] + cd /home/abuild/rpmbuild/BUILD/flac-1.5.0-build/BUILDROOT
[  291s] + mkdir -p .cfiles
[  291s] + pushd .cfiles
[  291s] /var/tmp/rpm-tmp.SodtsW: 41: pushd: not found
```

@bmwiedemann how does this break reproducible builds, please?

adrianschroeter:
I cherry-picked the mkbaselibs fix for now only, thanks.

PR #1064, opened 2025-04-03
Only treat mkosi.*.conf like files as mkosi

dirkmueller:
Otherwise it tries to build mkosi.spec as mkosi.


mlschroe:
Didn't you forget a `?` after your non-capturing regex group?

PR #1068, opened 2025-05-07
Reduce usage of bareword file handles

bmwiedemann:
Barewords in Perl are package level and may be interpreted as subroutine calls if a subroutine with a name colliding with the file handle's name is declared in the package.

This was one of the issues found via perlcritic by Radically Open Security.


mlschroe:
Not that I mind the change, as perl versions that don't support file handle autovivication are no longer supported by obs-build. But the change is actually a non-issue due to the `local` statement that kills the global namespace entry.

mlschroe:
(But I do not like the `open(my $fd)` form. Personal taste, I guess.

mlschroe:
But this needs to be changed so that lowercase variable names are used.

PR #1069, opened 2025-05-08
Add support for buildoption for productcompose via build config

adrianschroeter:


mlschroe:
That doesn't look right.

mlschroe:
Also it should be `productcompose-buildoption` or `productcompose.buildoption`.

adrianschroeter:
adapted as discussed

PR #1070, opened 2025-05-14
Disable selinuxfs mount during builds (bsc#1231252)

dirkmueller:
This is how mock does it. This only really changes something for SLFO where the behavior change is wanted. all other distros have a kernel that defaults to apparmor hence the change is a noop.

### DIFF
--- a/build-vm
+++ b/build-vm
@@ -646,9 +646,9 @@ vm_detect_2nd_stage() {
         # need to remount rw explicitly.
         mount -o remount,rw sysfs /sys
     fi
-    # mount selinuxfs
-    if ! test -e /sys/fs/selinux/status; then
-        mount -orw -n -tselinuxfs selinuxfs /sys/fs/selinux 2>/dev/null || :
+    # Hide selinux fstype (compatible with how mock does it)
+    if test -d /sys/fs/selinux; then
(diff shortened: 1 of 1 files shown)


Vogtinator:
FTR, for chroot builds I see:

```
│ ├─/tmp/osc-build-root/proc                            none                                          proc            rw,relatime
│ └─/tmp/osc-build-root/dev/pts                         none                                          devpts          rw,relatime,seclabel,gid=5,mode=620,ptmxmode=000
```

during preinstall and

```
│ ├─/tmp/osc-build-root/proc                            none                                          proc            rw,relatime
│ ├─/tmp/osc-build-root/sys                             none                                          sysfs           rw,relatime,seclabel
│ ├─/tmp/osc-build-root/dev/pts                         none                                          devpts          rw,relatime,seclabel,gid=5,mode=620,ptmxmode=666
│ └─/tmp/osc-build-root/dev/shm                         none                                          tmpfs           rw,relatime,seclabel,inode64
```

during builds, so `selinuxenabled` is false.

That means this PR makes `chroot` and VM builds more similar again as well

Vogtinator:
I tested chroot and VM builds of cockpit and kiwi-templates-Minimal:kvm-and-xen for SLFO and it all worked as expected

Vogtinator (approved):

PR #1072, opened 2025-05-16
fix compatibility with qemu 10.0 (fixes#1071)

seife:
qemu 10.0 has dropped the "-runas" option. Unfortunately the replacement "-run-with user=" was only introduced in qemu-9, which is not included in SLES15, so both variants need to be supported.
Try runtime detection of qemu and select appropriate options.


seife:
WARNING: I did not have the possibility to fix this (I don't have access to the OBS server where the tumbleweed runners are attached, my OBS instance does not have tumbleweed runners), so at least a cursory test might be advised. Shellcheck did at least not throw any new errors and did not complain about syntax :-)

PR #1086, opened 2025-07-09
qemu:m68k "virt" is supposed to be the faster emulation meanwhile

adrianschroeter:


andreas-schwab:
I don't know, there is no virt cpu in qemu 8.2.

adrianschroeter:
k, so my patch is wrong ... can/should we keep the -cpu m68000 and just add -M virt  ?

PR #1087, opened 2025-07-10
Additional configuration updates for build-vm-qemu for m68k

glaubitz:
This sets the proper console device and CPU for the virt machine type.


adrianschroeter:
@andreas-schwab any objection?

andreas-schwab:
Probably ok.

PR #1089, opened 2025-07-21
qemu-reg: Set the "fixed binary" flag to work inside chroots

Vogtinator:
This fixes kiwi builds with qemu-user.


andreas-schwab:
This causes a failure to unmount root during shutdown:

[  152s] ### VM INTERACTION START ###
[  152s] mount: /: mount point is busy.
[  152s]        dmesg(1) may have more information after failed mount system ca>
[  152s] Warning: 'remount -o ro /' failed. Triple Syncing..
[  152s] [  148.222047][    T1] sysrq: Power Off
[  152s] ### VM INTERACTION END ###

This only happens if the build didn't use a preinstall image, and thus
reinstalls qemu-linux-user during init_buildsystem, which unlinks the
preinstalled qemu-linux-user binary pinned down by the binfmt_misc
handler.

Vogtinator:
> If that's not possible, maybe init_buildsystem needs to redo the binfmt registrations.

Hm, that would actually break registrations in `%post` scripts like qemu-riscv64-btrfs. Not sure what the best fix is.

PR #1097, opened 2025-09-05
Remove elevator=noop from the command build vm kernel cmdline

crrodriguez:
Since the switch to blk-mq this does nothing.
the correct way is now either using an udev rule or echo "none" > /sys/block/*/queue/scheduler

It will be better if kernel-obs would default to none instead.

### DIFF
--- a/build-vm-kvm
+++ b/build-vm-kvm
@@ -360,7 +360,7 @@ vm_startup_kvm() {
         qemu_append="$qemu_append $vm_cmdline"
     else
         # Pick sensible defaults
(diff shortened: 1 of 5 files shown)


dirkmueller:
kernel-obs-build is just a repackaged kernel-default, so defaults can not be different. the files you're patching are also used for really old kernel versions, so the unconditional removal does not work there.

dirkmueller:
we'd have to implement the `echo none` trick then everywhere before removing it.

PR #1117, opened 2026-01-20
Improve capability parsing if not quoted

fcrozat:
When using capability in Dockerfile, if it is not quoted, obs-docker-support gets confused.

Both` zypper in -C "foo = 1.2"` and `zypper in -C foo = 1.2` are working fine on CLI or in a Dockerfile with podman but will fail inside OBS.

### DIFF
--- a/obs-docker-support
+++ b/obs-docker-support
@@ -127,6 +127,33 @@ zypper() {
     done
     case $cmd in
     in|install|rm|remove|up|update|if|info)
+	set -- "$@" "$sentinel"
+	local prev=""
+	local o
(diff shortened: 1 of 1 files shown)


mlschroe:
Hey, I don't think this is needed at all. I fixed the parser a while ago, the obs-docker-support should just pass the arguments to zypper and not mess with them.

fcrozat:
> Hey, I don't think this is needed at all. I fixed the parser a while ago, the obs-docker-support should just pass the arguments to zypper and not mess with them.

Indeed, closing.

PR #1123, opened 2026-02-13
debian: replace @BUILD_FLAVOR@ in parse and dsc source preparation

rjd15372:
This commit improves the support for multibuild when building debian-based packages.

* The `@BUILD_FLAVOR@` can now be used in the `.dsc` or `debian.control` file to correctly identify build dependencies.

* The `@BUILD_FLAVOR@` can also be used `debian/control` and `debian/rules` files, which are processed during build time but before the debian packaging process starts

### DIFF
--- a/Build/Deb.pm
+++ b/Build/Deb.pm
@@ -71,6 +71,7 @@ sub obsarch {
 
 sub parse {
   my ($bconf, $fn) = @_;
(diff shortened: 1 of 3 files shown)


rjd15372:
I'll open a new PR with just the `emptybuildflavor` attribute implementation.

adrianschroeter:
this got handled in #1124

PR #1139, opened 2026-04-23
podman: fix armv7 option

ggardet:
`--variant=v7` alone does not work. We also need to add `--arch=arm` or use `--platform=linux/arm/v7`


Vogtinator (changes requested):
Fails: `Error: unknown flag: --platform linux/arm/v7`

ggardet:
> Fails: `Error: unknown flag: --platform linux/arm/v7`

Replaced the space with a `=`

Vogtinator (approved):
Works locally!

PR #1141, opened 2026-04-29
Build::Docker: parse microdnf as a dnf-equivalent package manager

rjd15372:
The Dockerfile parser collects build dependencies by recognising package-manager invocations in `RUN` lines. microdnf shares dnf's `install`/`in` subcommand semantics, so route it through `cmd_dnf` alongside `yum` and `dnf`. Without this, `microdnf install <pkg>` commands in RedHat based images were silently ignored.

Extend `t/parse_docker.t` with a `microdnf install` case to cover the new dispatch.


mlschroe:
Thanks!

PR #1147, opened 2026-05-27
also use cmdline.obs.guest

bugfinder:


build-vm-kvm
```diff
@@ -207,6 +208,9 @@ vm_verify_options_kvm() {
     if test -z "$vm_kernel" ; then
 	vm_kernel=/boot/vmlinuz
 	test -e /boot/kernel.obs.guest && vm_kernel=/boot/kernel.obs.guest
+	if test -z "$vm_cmdline" ; then
```
Vogtinator, line 211: IMO this needs to be done unconditionally like above to ensure that if kernel.obs.guest is used, cmdline.obs.guest is also used?

bugfinder: the only other place setting it is if a kernel-obs-build package is installed in the VM_ROOT, read if the .build.cmdline.kvm is there, and then this is not the fallback-to-host-provided-kernel case.

PR #1148, opened 2026-05-30
 Avoid process substitution in the init_buildsystem script

mcepl:
Bash process substitution requires a working /dev/fd, which may
 be missing in nested chroot/container builds. Use regular
 pipeline/tempfile input instead in restricted build roots.

 References: https://github.com/openSUSE/brp-check-suse/issues/54

### DIFF
--- a/init_buildsystem
+++ b/init_buildsystem
@@ -389,6 +389,14 @@ create_devs() {
 DEVLIST
 }
 
+create_dev_fd_links() {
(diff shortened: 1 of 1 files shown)


adrianschroeter:
based on personal discussions, this might be fixed already. reopen if not.

mcepl:
This is absolutely not resolved, see the discussion on https://github.com/openSUSE/brp-check-suse/pull/55, which seems to be leading nowhere though.

PR #1149, opened 2026-06-02
Add support for comparing deb packages

bluca:
build-compare just gained support for comparing the result of deb builds too, so add support for it, following the rpm model

### DIFF
--- a/build-recipe
+++ b/build-recipe
@@ -64,6 +64,10 @@ recipe_resultdirs () {
     recipe_resultdirs_$BUILDTYPE "$@"
 }
 
+recipe_compare_oldpackages () {
+    recipe_compare_oldpackages_$BUILDTYPE "$@"
+}
+
 recipe_cleanup () {
     recipe_cleanup_$BUILDTYPE "$@"
 }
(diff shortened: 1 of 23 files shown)


build-recipe
```diff
@@ -64,6 +64,15 @@ recipe_resultdirs () {
     recipe_resultdirs_$BUILDTYPE "$@"
 }
 
+recipe_compare_oldpackages () {
+    # delegate to a buildtype-specific implementation if one exists.
+    # buildtypes that have no useful "same result" detection (e.g. image
+    # builds) simply leave the function undefined and we no-op here.
+    if declare -F "recipe_compare_oldpackages_$BUILDTYPE" > /dev/null ; then
+	"recipe_compare_oldpackages_$BUILDTYPE" "$@"
+    fi
```
adrianschroeter: please avoid this check, implement a no-op in the build-receipe-* implementations instead.

adrianschroeter (changes requested):

build-recipe
bluca: ok, done

PR #1151, opened 2026-06-05
generate_sbom: fix runtime errors on old distributions

adrianschroeter:
We run generate_sbom there without BUILD_DIR set, but using perl -I /.build directly

However, generate_sbom is trying to use /usr/lib/build in this situation for it's modules which leads to runtime errors (opts->{'type'} is not set leading to handle everything as oci container).

Instead let's use the path of generate_sbom as perl module directory if BUILD_DIR is not set.

### DIFF
--- a/generate_sbom
+++ b/generate_sbom
@@ -21,7 +21,12 @@
 ################################################################
 
 BEGIN {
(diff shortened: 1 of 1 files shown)


mlschroe:
Who is "we"?

PR #1158, opened 2026-07-01
fix: build-receipe-spec: parsing %generate_buildrequires with $BUILD_USER

SalimTerryLi:
Both static and dynamic BuildRequires were collected with root user before, resulting `%{_topdir}` pointing to wrong place. This breaks `%include` within spec file.

This PR attempts to make the process happen under `$BUILD_USER` so that rpmspec tool can refer to the correct source dir.

Seems that `.rpmmacros` and `.rpmrc` exist both in `/root` and `/home/$BUILD_USER` with identical content, I didn' t catch the reason that it was done previously by root user. Is it specially designed to do so?

This PR is tested on our OBS instance as well as `pbuild`. However `osc` does not support `%generate_buildrequires` well.


mlschroe:
Thanks!

Issue #1161, opened 2026-07-22
[bug]: Docker RUN instructions not correctly parsed when `--mount` is used


SchoolGuy:
During my work on the Orthos 2 Container, I was wondering why OBS didn't correctly replace my ARG inside the `RUN` instruction. @Vogtinator was so kind as to explain to me that there is an outstanding issue where the parsing of `RUN` instructions is not covering all edge cases.

Not working:

```dockerfile
ARG PROJECT="production"
RUN --mount=type=secret,id=SCCcredentials,target=/etc/zypp/credentials.d/SCCcredentials,required=false \
    zypper --non-interactive addrepo --refresh "https://download.opensuse.org/repositories/systemsmanagement:orthos2:${PROJECT}/15.7/" "Orthos 2 ${PROJECT}" && \
    zypper --non-interactive --gpg-auto-import-keys refresh
```

Workaround:

```dockerfile
ARG PROJECT="production"
RUN --mount=type=secret,id=SCCcredentials,target=/etc/zypp/credentials.d/SCCcredentials,required=false true && \
    zypper --non-interactive addrepo --refresh "https://download.opensuse.org/repositories/systemsmanagement:orthos2:${PROJECT}/15.7/" "Orthos 2 ${PROJECT}" && \
```

The example was shortened inside this issue. The full example can be seen inside the Orthos 2 PR https://github.com/openSUSE/orthos2/pull/415. The reason for this long run instruction is that I can use the same Dockerfile locally and inside OBS. This is used to build the following two Docker Images to be found in the following OBS project: https://build.opensuse.org/project/show/systemsmanagement:orthos2

VithHackie:
"Hi! I've been looking into how obs-build parses Dockerfile RUN instructions and I'd like to take a shot at fixing this line-continuation / ARG substitution edge case. Could you assign this to me?"

mlschroe:
Just open a pull request. But it's really just stripping `--mount=` and `--network=` from the RUN line, e.g. by adding:
```
1 while $line =~ s/^--(?:mount|network)=\S*\s+//;
```

mlschroe:
Fixed now.

PR #1163, opened 2026-07-27
Fix: Strip BuildKit mount and network flags from RUN instructions in …

VithHackie:
## Title: 
Fix issue with --mount flags in Dockerfile RUN commands

### This fixes issue #1161.

When a Dockerfile has a RUN command with flags like --mount or --network, and uses a backslash \ to continue onto the next line, the patchdockerfile script was getting confused. It was basically ignoring the rest of the command and failing to do the ARG replacements.

## How I fixed it
-> I used the Perl regex that @Vogtinator kindly suggested in the issue comments!

## Here is what the code does now:

- > It temporarily removes the RUN keyword from the start of the line.

- > It uses the regex to vacuum up the --mount= and --network= flags.

- > It pastes RUN back onto the front.

This cleanly removes the unsupported flags while keeping the line continuation \ perfectly intact so the parser can read the next line.

## Testing
- > [x] Verified the regex successfully strips --mount and --network flags.

- > [x] Tested locally against the failing Dockerfile syntax provided in the original issue to ensure line continuations and ARG substitutions now process correctly.

***Thanks for the guidance in the issue thread! Let me know if everything looks okay or if I need to make any changes.***

### DIFF
--- a/patchdockerfile
+++ b/patchdockerfile
@@ -54,6 +54,11 @@ for my $s (@sections) {
   my ($install_idx, $uninstall_idx, $from_scratch);
   my $idx = 0;
   for my $line (@$s) {
+    # --- NEW CODE TO TACKLE SIDE CASES WITH --mount ---
+    if ($line =~ s/^([rR][uU][nN]\s+)//) {
+        1 while $line =~ s/^--(?:mount|network)=\S*\s+//;
+        $line = $1 . $line;
+    }
     if ($line =~ /^[uU][sS][eE][rR]\s+root[\s\n]/) {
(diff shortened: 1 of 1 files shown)


mlschroe:
I don't think this is the correct place. The `patchdockerfile` helper is just inserting the obs-docker-support extra statements.

What you need to patch is the Dockerfile parser in `Build/Docker.pm`. Look for the `} elsif ($cmd eq 'RUN') {` line.

mlschroe:
No response. Now fixed in commit 6fcaff74ba83130a85b44f6fb2202dfaab23c022

PR #1166, opened 2026-09-02
copy_sources: normalize permissions of copied sources

bmwiedemann:
Copied sources now always end up as 644 (755 for directories and executables) to ensure consistent permissions in the generated src.rpm.

This patch was done while working on reproducible builds for openSUSE.


adrianschroeter:
why shouldn't it be reproducible without?

Our legacy system is not handling permissions usually, but with git we will get it. 

We would break then the permissions from git if we overwrite them here.

bmwiedemann:
I had a problem with `osc build` where my checkout was done with `umask 2` instead of the usual `umask 22` and it produced different binaries from the OBS builds.

PR #1167, opened 2026-09-02
generate_sbom: Turn whitespace off by default when filelist generatio…

dirkmueller:
…n is on

This saves about 20% on disk size and transfer size for rekor uploads.

### DIFF
--- a/generate_sbom
+++ b/generate_sbom
@@ -1103,11 +1103,11 @@ sub spdx3_encode_disjunctive_license {
 sub spdx3_encode_tokenized_license {
   my ($doc, $n) = @_;
   my @n = ('START', @$n);
-  my $l = ''; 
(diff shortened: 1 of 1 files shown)


adrianschroeter:
can we name the switch --compact or something some other positive term to give a reason for it's existens? ;)

dirkmueller:
> can we name the switch --compact or something some other positive term to give a reason for it's existens? ;)

are you okay wtih "--no-pretty" ? that would match the name in createrepo_c

adrianschroeter:
k, if this is kind of a standard ;)

dirkmueller:
> k, if this is kind of a standard ;)

It's not really a standard (I have invented that name when I contributed the change to createrepo). but it is always good to be consistent with your own standards ;)

PR #1173, opened 2026-09-15
Honor /usr/lib/build/rpmdb-source-date-epoch when installing rpms

bmwiedemann:
Honor /usr/lib/build/rpmdb-source-date-epoch when installing the build root

rpm records RPMTAG_INSTALLTIME and RPMTAG_INSTALLTID from the wall clock, so the rpm database of a build root is different on every build. That is harmless for ordinary packages, but appliance and container builds copy the database into the artifact they ship, and it makes those unreproducible. rpm >= 4.18 uses SOURCE_DATE_EPOCH instead of the wall clock for both tags when it is set.

obs-build is shared by every distribution built on an OBS instance, so this must not be switched on unconditionally. Instead a distribution opts in by shipping /usr/lib/build/rpmdb-source-date-epoch in one of the packages on its preinstall list. Those packages are unpacked before the first rpm -U into the build root, in chroot and VM builds alike, and /usr/lib/build is already the place for package-provided hooks.

An empty file selects epoch 0, so that the transaction id is 0 and the install times are numbered in package order. Otherwise the file must contain the epoch as a single integer; anything else is ignored with a warning. The environment is only set for the rpm calls that install packages, not for the build itself, which uses its own SOURCE_DATE_EPOCH derived from the recipe.

Closes #1164


Vogtinator:
IMO this should be a buildflag instead

